PR #261, opened 2024-10-26
UI: Refine leg filtering and add debug logging in TripResponseMapper

ksharma-xyz:
### TL;DR

If the first leg is a walking leg, then it does not have any stop sequence and especially for Metro and Train. In those cases the walking leg was getting filtered out from  `val legs = journey.getFilteredValidLegs()` and not getting mapped ot Model in `private fun TripResponse.Leg.toUiModel()`.

Since only Public Transport legs have stopSequence it can be safely removed from the filter logic.

### What changed?

- Modified `getFilteredValidLegs()` function to only filter legs based on transportation, removing the check for `stopSequence`.
- Added detailed debug logging in the `toUiModel()` function to provide more information about each leg during the mapping process.

### Why make this change?

This change aims to improve the accuracy of leg filtering and provide better debugging capabilities. By removing the `stopSequence` check, we ensure that valid legs with transportation information are not accidentally filtered out. The added logging will help in troubleshooting and understanding the leg mapping process more effectively.

### DIFF
--- a/feature/trip-planner/ui/src/main/kotlin/xyz/ksharma/krail/trip/planner/ui/timetable/business/TripResponseMapper.kt
+++ b/feature/trip-planner/ui/src/main/kotlin/xyz/ksharma/krail/trip/planner/ui/timetable/business/TripResponseMapper.kt
@@ -68,9 +68,7 @@ private fun TripResponse.Leg?.getDepartureTime() =
 private fun TripResponse.Leg?.getArrivalTime() =
     this?.destination?.arrivalTimeEstimated ?: this?.destination?.arrivalTimePlanned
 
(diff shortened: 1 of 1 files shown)


ksharma-xyz:
* **#261** <a href="https://app.graphite.dev/github/pr/ksharma-xyz/Krail/261?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @ksharma-xyz and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

PR #380, opened 2024-11-26
Fix display name for Leg

ksharma-xyz:
### TL;DR
Updated journey card display text to show destination name instead of transportation description

### What changed?
Modified the `toUiModel()` function in `TripResponseMapper.kt` to use `transportation?.destination?.name` instead of `transportation?.description` for the `displayText` property

### How to test?
1. Open the trip planner
2. Search for and select a journey
3. Verify that journey cards now display the destination name
4. Compare with previous behavior where transportation description was shown

### Why make this change?
Displaying the destination name provides more relevant information to users when viewing journey details, making it easier to identify where each leg of their trip ends


ksharma-xyz:
### Merge activity

* **Nov 26, 3:34 AM EST**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/ksharma-xyz/Krail/380).

PR #695, opened 2025-04-04
Refactor: Replace DisplayWithDensityCheck with isFontScaleLessThanThreshold function

ksharma-xyz:
### TL;DR

Refactored accessibility handling for font scaling in the trip planner UI.

### What changed?

- Replaced `DisplayWithDensityCheck` composable with a new `isFontScaleLessThanThreshold` function that returns a boolean instead of conditionally displaying content
- Updated the `DefaultJourneyCardContent` to use the new function with additional logic to only show transport mode list when there are fewer than 4 items or the font scale is below the threshold
- Removed unnecessary `@OptIn` annotations from `DefaultJourneyCardContent`

### How to test?

1. Test the UI with different font scale settings (both below and above 1.8x)
2. Verify that transport mode lists with fewer than 4 items always display regardless of font scale
3. Verify that transport mode lists with 4+ items only display when font scale is below 1.8x
4. Ensure no regressions in journey card display

### Why make this change?

This change improves accessibility by providing more granular control over what UI elements are displayed based on font scale. The new approach allows for more flexible conditional rendering based on both content size and font scale, rather than just hiding content based on font scale alone.

### DIFF
--- a/feature/trip-planner/ui/src/commonMain/kotlin/xyz/ksharma/krail/trip/planner/ui/components/A11yExt.kt
+++ b/feature/trip-planner/ui/src/commonMain/kotlin/xyz/ksharma/krail/trip/planner/ui/components/A11yExt.kt
@@ -3,13 +3,8 @@ package xyz.ksharma.krail.trip.planner.ui.components
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.platform.LocalDensity
 
-/**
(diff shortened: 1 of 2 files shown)


ksharma-xyz:
### Merge activity

* **Apr 4, 7:18 AM EDT**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/ksharma-xyz/KRAIL/695).

PR #759, opened 2025-05-10
BUG: Improve journey caching and add detailed logging for trip planner

ksharma-xyz:
## Problem

Previously, the cache merge logic only preserved:  
- All journeys from the latest API response.
- Up to 2 "started" journeys from cache (not in API and not expired).

This meant that if a journey was not in the API and not considered "started" (e.g., a future journey), it was dropped from the cache and UI. As a result, journeys like the 2:59 PM trip disappeared from the UI when the API stopped returning them, even if they were still relevant (not expired).  

**Note,** API sometimes does not provide the latest upcoming journey as part of response.  e.g time now is 2:00 pm then API does not give us 2:04 train, it gives 2:08 and 2:10. however the train is not cancelled and still in future (upcoming), so to safeguard against this issue we need to cache and display it. Sometimes the train may actualaly be cancelled in which case this will be an issue though.  

## Solution

The cache filter logic is now updated to:  
- Split cached journeys not in the latest API response into two groups:
  - startedJourneys: Up to 2 most recent journeys that have already started (origin time <= now).
  - upcomingJourney: The single earliest journey that is scheduled to start in the future (origin time > now).
   **NOTE:** We need to keep only one upcoming Journey, because API sometimes does not provide the latest upcoming journey as part of response.  
  
- These are combined into eligibleCacheJourneys (startedJourneys + upcomingJourney), which are then merged with the latest API journeys for display.

### Purpose

This approach ensures that:
- Users always see up to 2 recently started journeys, even if the API omits them, so recent/past trips remain visible for a short period.
- Only the most relevant missing upcoming journey (the one closest to now) is kept, preventing the cache from growing with less relevant future trips.
- The UI remains consistent and focused on the most important journeys, improving user experience and cache efficiency.

## Summary of Changes
- Changed the cache filter to select up to 2 most recent, not-expired journeys from cache that are not in the latest API response, regardless of their "started" status.

- This prevents relevant journeys from disappearing from the UI when the API omits them.

This fix improves the user experience by ensuring that recent and upcoming journeys are reliably shown, even if temporarily missing from the API response.

### DIFF
--- a/feature/trip-planner/network/src/androidMain/kotlin/xyz/ksharma/krail/trip/planner/network/api/service/HttpClient.kt
+++ b/feature/trip-planner/network/src/androidMain/kotlin/xyz/ksharma/krail/trip/planner/network/api/service/HttpClient.kt
@@ -4,7 +4,6 @@ import io.ktor.client.HttpClient
 import io.ktor.client.engine.okhttp.OkHttp
 import io.ktor.client.plugins.contentnegotiation.ContentNegotiation
 import io.ktor.client.plugins.defaultRequest
-import io.ktor.client.plugins.logging.ANDROID
 import io.ktor.client.plugins.logging.LogLevel
 import io.ktor.client.plugins.logging.Logger
 import io.ktor.client.plugins.logging.Logging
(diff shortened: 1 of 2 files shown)


ksharma-xyz:
* **#759** <a href="https://app.graphite.dev/github/pr/ksharma-xyz/KRAIL/759?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/ksharma-xyz/KRAIL/759?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #836, opened 2025-06-17
Save stopname along with other facility name data and then use stop name to display rather than falling back on tsn

ksharma-xyz:
Save stopname along with other facility name data and then use stop name to display rather than falling back on tsn

this ahndles edge case, where tsn not present in stops list so we dont know stop name, therefore imp to save stopname in table

Remove unused events


ksharma-xyz:
### Merge activity

* **Jun 17, 10:01 AM UTC**: @ksharma-xyz merged this pull request with [Graphite](https://app.graphite.dev/github/pr/ksharma-xyz/KRAIL/836).

PR #1107, opened 2025-09-13
Implement responsive layout for JourneyCard information row

ksharma-xyz:
# Improved JourneyCard layout with responsive info row

Refactored the `ResponsiveJourneyInfoRow` component to handle three distinct layout scenarios:

1. Single row layout when all elements fit: destination | clock+walk ... deviation
2. Two-row default layout: Row1 destination; Row2 clock+walk | ... deviation
3. Fallback layout when Row2 can't fit clock+walk+deviation: Row1 destination+clock; Row2 walk | ... deviation

## Changes in `ResponsiveJourneyInfoRow`

\- Replaced `FlowRow` with a `SubcomposeLayout` to measure and place items precisely.
\- Measured four atomic parts separately: destination, clock, walk, deviation.
\- One-row layout when all fit: destination at start, deviation at end, clock+walk in the middle.
\- Two-row default when not fitting: row 1 destination only; row 2 clock+walk at start, deviation at end.
\- Added third fallback: if row 2 cannot fit clock+walk+deviation, move clock next to destination in row 1; put walk at start of row 2 and deviation at end.

The implementation uses `SubcomposeLayout` to measure each element individually and apply the appropriate layout strategy based on available space, ensuring optimal information display across different screen sizes and content lengths.

## Visuals

https://github.com/user-attachments/assets/f32c2624-e585-40ce-8384-947e0a5fffa1

### DIFF
--- a/feature/trip-planner/ui/src/commonMain/kotlin/xyz/ksharma/krail/trip/planner/ui/components/JourneyCard.kt
+++ b/feature/trip-planner/ui/src/commonMain/kotlin/xyz/ksharma/krail/trip/planner/ui/components/JourneyCard.kt
@@ -33,6 +33,7 @@ import androidx.compose.ui.graphics.Brush
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.ColorFilter
 import androidx.compose.ui.graphics.painter.Painter
+import androidx.compose.ui.layout.SubcomposeLayout
 import androidx.compose.ui.platform.LocalDensity
 import androidx.compose.ui.semantics.Role
 import androidx.compose.ui.semantics.semantics
@@ -197,8 +198,10 @@ private fun DepartureDeviationIndicator(
     val (dotColor, label) = when (deviation) {
         is TimeTableState.JourneyCardInfo.DepartureDeviation.Late ->
             KrailTheme.colors.deviationLate to deviation.text
+
         is TimeTableState.JourneyCardInfo.DepartureDeviation.Early ->
             KrailTheme.colors.deviationEarly to deviation.text
+
         TimeTableState.JourneyCardInfo.DepartureDeviation.OnTime ->
             KrailTheme.colors.deviationOnTime to "On time"
     }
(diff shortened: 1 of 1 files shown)


ksharma-xyz:
## High‑level: what are the phases in Compose
- Composition: Compose runs your composable functions and builds a tree of UI nodes. Think “what to show.”
- Layout (measure and place): Parents give Constraints to children, children return Placeables with a size, and the parent decides where to place them. Think “how big and where.”
- Drawing: The final pixels are drawn. Think “paint.”

## What makes `SubcomposeLayout` special
- It lets you compose children later, inside the measure pass, using subcompose(slotId) { … }. That means you can measure some content first, compute available space, and then decide what else to compose and how to lay it out. Regular layouts compose all children up front and cannot easily do this kind of adaptive measurement.
- It caches those subcompositions by slotId. On subsequent layouts, calling subcompose with the same slotId reuses the composition and only remeasures, unless the inputs changed and require recomposition. This is why it stays performant for small numbers of slots.

## The core objects in the layout pass
- Constraints: A rectangle of allowed sizes (minWidth, maxWidth, minHeight, maxHeight) expressed in pixels. Parents must measure children with some Constraints.
- Measurable: A handle representing a child to be measured. You obtain these from subcompose.
- Placeable: The result of measuring a Measurable. It has width and height and can be placed in the parent with placeRelative(x, y). placeRelative respects RTL automatically.

## What your SubcomposeLayout does step by step
Subcompose four slots: dest, clock, walk, dev. Each call returns a list of Measurable because a slot can emit multiple nodes.
Measure each list with constraints to get Placeables and their sizes. Now you know destW, clockW, walkW, devW, etc.
Decide the layout variant: 3.1 One row fits if clockW + walkW fits between dest on the left and dev on the right. 3.2 Otherwise two rows default: row 1 only dest; row 2 has clock and walk on the left and dev on the right. 3.3 Fallback split: if row 2 still cannot fit clock + walk + dev, move clock up next to dest on row 1; keep walk at start of row 2 and dev at the end.
Call layout(parentWidth, parentHeight) and place the Placeables: 4.1 Left‑to‑right placement is done by accumulating an x cursor for dest, then clock, then walk. 4.2 dev is right‑aligned by placing it at devX = maxW - devW. 4.3 For the second row, use y offset equal to the first row’s height.
Because you use placeRelative, it will flip automatically for RTL locales, keeping “start” and “end” semantics correct.

## Why this is performant here
- You compose only four small slots. Subcompose adds a tiny overhead but gives you precise, single‑pass control. This avoids the trial‑and‑error multi‑measure that generic wrapping containers may do.
- Reuse via slotId means future layouts typically remeasure without recomposing. If only width changes (e.g., device rotation), -Compose can often skip composition and just redo measure and place.
-You measure each slot once with clear constraints, and you do not use weight, so children do not expand unexpectedly and do not force remeasures.

## Subcompose details worth knowing
- Each subcompose call can emit zero, one, or many Measurables. Measuring returns Placeables. Always sum or max the sizes you need; one slot might contribute multiple nodes.
- You can pass tighter Constraints to measure children more efficiently. In your code you measured with the full constraints, which is fine; you could optimize by measuring mid content with maxWidth equal to the remaining space, but you already compute fit from widths and choose a layout variant explicitly.
- Intrinsics: parents may ask children for “intrinsic” sizes. SubcomposeLayout implements intrinsic measurements by performing a lightweight subcomposition as needed. Intrinsics can be costly; avoid relying on them in hot paths if possible.

## Why placeRelative versus place
placeRelative(x, y) mirrors horizontally in RTL, preserving “start” and “end” semantics. Since your rules are “destination at start” and “deviation at end,” placeRelative is the correct choice.

PR #1489, opened 2026-04-18
Derive iOS build number from ASC API with monotonic GitHub floor fallback

ksharma-xyz:
### TL;DR

Replaces the static GitHub variable-based iOS build number increment with a dynamic strategy that queries App Store Connect (ASC) for the latest accepted build number, ensuring CI never uploads a duplicate or regressed `CFBundleVersion`.

### What changed?

- Added `iosApp/scripts/ios_next_build_number.py`, a Python script that computes the next build number using `next = max(asc_latest, github_floor) + 1`. It authenticates with ASC via a JWT (using `PyJWT` and `cryptography`) and pages through all builds to find the highest version Apple has accepted. The GitHub variable acts as a monotonic floor in case the ASC API returns a stale value.
- The "Set iOS build number" CI step now runs this script inside a Python venv (to avoid Homebrew PEP 668 conflicts) and exports `IOS_BUILD_NUMBER` via `$GITHUB_ENV`. The GitHub variable write-back is retained as a best-effort audit trail.
- A guard was added to the "Build iOS App for Release" step that aborts the build if `IOS_BUILD_NUMBER` is not a valid positive integer, preventing an IPA from being stamped with a stale build number.
- `CURRENT_PROJECT_VERSION` and `MARKETING_VERSION` were moved from `project.pbxproj` into `Config.xcconfig` as the single source of truth. The Fastfile now reads `MARKETING_VERSION` from `Config.xcconfig` and passes both values via `xcargs` at archive time, which is required when `GENERATE_INFOPLIST_FILE = YES`.
- `PRODUCT_BUNDLE_IDENTIFIER` and `PRODUCT_NAME` are now defined in `Config.xcconfig` and referenced via `$(...)` in the project file, replacing the old `${BUNDLE_ID}` / `${APP_NAME}` variables.
- The `brew install swiftlint` step was removed from the CI workflow.
- Several GitHub Actions were pinned to stable major versions (`actions/checkout@v4`, `setup-java@v4`, `setup-gradle@v4`, `upload-artifact@v4`, `import-codesign-certs@v3`, `download-provisioning-profiles@v3`).

### How to test?

Trigger the `distribute-testflight` workflow and verify:
1. The "Set iOS build number" step prints both the ASC latest value and the GitHub floor, and the computed next build number is greater than both.
2. The IPA is stamped with the correct `CFBundleVersion` and `CFBundleShortVersionString` matching `Config.xcconfig`.
3. TestFlight accepts the upload without rejecting the build number.
4. The `IOS_BUILD_NUMBER` repository variable is updated to the new value after a successful run.

### Why make this change?

The previous approach incremented a GitHub Actions variable by 1 on each run, which had no awareness of manual uploads or out-of-band builds. Apple rejects uploads where `CFBundleVersion` is not strictly greater than the last accepted build, so any manual upload would cause the next CI run to fail. Querying ASC directly makes the build number monotonically correct regardless of how builds are submitted. Moving version metadata into `Config.xcconfig` also eliminates duplication between the project file and CI configuration.


ksharma-xyz:
* **#1489** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1489?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1489?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1518, opened 2026-04-24
fix: encode excluded product classes in deep link and auto-switch trips on new deep link tap

ksharma-xyz:
### TL;DR

Tapping a deep link to a different trip now automatically clears the previous tracking session and starts the new one, instead of showing an "Already Tracking" interstitial screen.

### What changed?

- When a deep link is tapped for a different trip while one is already being tracked, the old tracking session is stopped immediately and the new trip's `Prompt` state is shown — no confirmation dialog required.
- The `AlreadyTracking` UI state and its associated composable (`AlreadyTrackingContent`) have been removed as they are now unreachable.
- `excludedProductClasses` is now encoded into the deep link so that when a tracked trip is refreshed via the TrackTrip API, the same transport mode filters that were active when the journey was planned are re-applied. This prevents, for example, a bus-only journey from returning train results.
- Old deep links without the `excl` field decode gracefully with an empty list (no exclusions applied).

### How to test?

1. Open the timetable for a bus-only route (deselect all modes except bus), tap **Track** on a journey, and confirm tracking starts.
2. While tracking that bus journey, tap a deep link for a different trip. Verify the old tracking is cleared automatically and the new trip's prompt screen is shown — no "Already on another trip" screen should appear.
3. Confirm that the TrackTrip refresh correctly returns bus journeys (not trains) for a bus-only tracked trip.
4. Decode an old deep link (without the `excl` field) and verify it loads without errors, applying no product class exclusions.

### Why make this change?

Tapping a deep link is an explicit user action to switch trips — forcing users through a confirmation screen adds unnecessary friction. Automatically clearing the old session and starting the new one is the more intuitive behaviour. Additionally, without encoding the excluded product classes into the deep link, the TrackTrip API would default to returning train journeys even when the user had filtered to a different mode, causing incorrect results during trip tracking.

### DIFF
--- a/feature/track/state/src/commonMain/kotlin/xyz/ksharma/krail/feature/track/TripDeepLink.kt
+++ b/feature/track/state/src/commonMain/kotlin/xyz/ksharma/krail/feature/track/TripDeepLink.kt
@@ -13,6 +13,14 @@ data class TripDeepLink(
     @SerialName("dep") val departureUtcDateTime: String,
     /** One entry per TransportLeg in the shared journey */
     @SerialName("legs") val legs: List<DeepLinkLeg>,
+    /**
+     * TfNSW product classes excluded when this journey was planned (e.g. [1] when the user
+     * filtered to bus-only). The TrackTrip API call must apply the same exclusion so it returns
+     * bus journeys instead of defaulting to trains.
+     *
+     * Absent from old deep links → decoded as [emptyList] (no exclusions applied).
+     */
+    @SerialName("excl") val excludedProductClasses: List<Int> = emptyList(),
 ) {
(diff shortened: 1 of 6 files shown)


ksharma-xyz:
* **#1518** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1518?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1518?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1521, opened 2026-04-28
feat: add StopLabel model and stop-labels field to state

ksharma-xyz:
feat: add StopLabel model and stop-labels field to state

Introduces StopLabel with emoji/label/stopId/stopName, isSet helper, and
toStopItem() conversion. Adds StopLabelPillTapped and AddStopLabelTapped
events. Seeds SavedTripsState with Home/Work defaults.

feat: add StopLabelPillRow component

LazyRow of set/unset/add pills using Taj chip tokens. Set pills use
theme colour background; unset and add pills use a thin onSurface border.
Includes 3 @Preview functions.

feat: redesign SearchStopRow with collapsible pill and From reveal animation

Collapsed state shows a centred "+ Plan a trip" pill. Expanded state
slides up via AnimatedContent with spring physics. From field hides
initially when isFromHighlighted=true and reveals after 350 ms via
expandVertically+fadeIn. Pulsing border alpha (0.4→1.0, 800 ms) cues
the user to fill From. Includes 5 @Preview functions.

feat: wire label pills and redesigned search row into SavedTripsScreen

StopLabelPillRow appears above the trip list. Tapping a set pill
pre-fills the To stop and expands+highlights the search row. The
ViewModel handles StopLabelPillTapped by calling
stopResultsManager.setSelectedToStop. Includes 3 @Preview functions.

feat: redesign LocationPermissionBanner as compact inverted pill

Replaces the verbose Column layout with a single-row pill using
onSurface background and surface text (inverted colours). Content is
"Location off — Open Settings ×" with a dismiss callback defaulting to
no-op so existing call sites are unchanged.

fix: resolve detekt violations in SearchStopRow and SavedTripsScreen

Fix import ordering (auto-corrected) in SearchStopRow. Remove unused
onReverseButtonClick parameter from SavedTripsScreen — the lambda
already calls onEvent(ReverseStopClick) directly. Remove the matching
arg from SavedTripsEntry.

feat: navigate to SearchStop when tapping an unset label pill

Adds LABEL to SearchStopFieldType and labelKey to SearchStopRoute /
StopSelectedResult. Tapping Home/Work (unset) navigates to SearchStop
with fieldType=LABEL; the result comes back as StopLabelAssigned which
updates that label's stop in SavedTripsState.

feat: map-first SearchStop with one-time options sheet and tap-to-select hint

SearchStopScreen now defaults to map view (showMap=true). First open
auto-pops MapOptionsBottomSheet and marks it seen via SandookPreferences
so it never auto-pops again. MapOptionsBottomSheet gains a hint line
"Tap any stop on the map to select it". List stays visible as a
placeholder while the map initialises to avoid a blank frame.

feat: label pill sheets — use-as from/to and add new label flow

Tapping a set label pill opens StopUseAsBottomSheet with "Going here" /
"Starting from here" buttons. Tapping "+" opens AddLabelBottomSheet with
suggestion chips and a name field, then navigates to SearchStop to pick
the stop; on return the new label is created from PendingNewLabel state.

refactor: reuse map StopDetailsBottomSheet for label pill "use as" action

Replace the custom StopUseAsBottomSheet with the existing map stop
details sheet (departure board + transport modes), parameterised via
an actionContent slot. Tapping a set label pill now shows the same rich
stop sheet with "Going here" / "Starting from here" buttons instead of
"Select Stop". StopUseAsBottomSheet.kt deleted.

feat: dynamic placeholder, select-on-map button, remove map pill in SearchStop

- Placeholder text is now context-aware: "Choose starting point" (FROM),
  "Choose destination" (TO), "Choose a stop" (LABEL)
- Map toggle pill removed from SearchTopBar — users switch to map via list
- "Select on map" button (location icon + theme colour) pinned as the
  first item in both Recent and Results lists; clicking it opens the map
  view and initialises it if needed

feat: conditional SearchStopRow and smoother expand animation

Show collapsed pill only when content is rich (≥2 trips + Park & Ride +
≥2 departure boards); otherwise SearchStopRow starts expanded. Replaced
spring/EaseOutBounce animations with tween(600ms, FastOutSlowInEasing)
for a slower, smoother reveal with no bounce.

fix: klickable lambda and StopActionButton import compile errors

feat: persist stop labels to SQLDelight DB (migration 8)

- Add StopLabels.sq table (label, emoji, stop_id, stop_name, sort_order)
- Migration 8 creates table and seeds Home/Work defaults for upgrades
- Fresh installs seeded in ViewModel when DB is empty on first observe
- Sandook interface + RealSandook implement observe/upsert/update/delete
- SavedTripsViewModel observes DB reactively; label assignments now persist

feat: stop-first add-label UX — pick stop then name it

- Tap "+" navigates directly to SearchStop (no label key yet)
- On stop selection return, naming bottom sheet appears with stop
  name as subtitle, emoji chips, text field, and "Save" button
- Dismiss without saving creates nothing (safeguard)
- AddLabelBottomSheet repurposed: takes stopName, onSave callback,
  imePadding for keyboard, no "Pick a stop" detour
- NewLabelCreated event writes directly to DB via sandook.upsertStopLabel
- Existing label update flow (unset label pill tap) unchanged

feat: label pill icons, onSurface text, chip sync, animated naming sheet

- "Select on map" text uses onSurface color (icon stays theme-tinted)
- Label pills use ic_location icon instead of emoji (clearer on themed bg)
- Chip selection syncs to text field via key(selectedSuggestionIndex)
- AddLabelBottomSheet: "What do you call [stop]?" heading with stop name
  in a themed pill chip; live animated label preview updates as user types

fix: onSurface icon, nickname heading, config-change persistence, PT note

- SelectOnMapItem icon tinted onSurface to match text color
- AddLabelBottomSheet heading: "Give a nickname to [stop chip]"
- pendingLabelStop stored as rememberSaveable JSON string — survives rotation
- SearchStopScreen: "Only public transport stops are shown." note at list end

feat: move stop label pills to SearchStopScreen; add outlineSubtle token

Label quick-pick shortcuts (Home, Work, etc.) now appear at the top of
SearchStopScreen — tapping a set label immediately selects that stop.
Removed labels from SavedTripsScreen, AddLabelBottomSheet flow, and all
related state/events. SavedTripsViewModel seeds default labels once on
startup; SearchStopViewModel observes and exposes them to the UI.

Added `outlineSubtle` color token to KrailColors (light 0x33010101 /
dark 0x33FCF6F1) and replaced all onSurface.copy(alpha=0.2–0.3f) border
and divider uses. Replaced remaining onSurface.copy(alpha=0.5–0.7f) text
uses with the existing softLabel token.

Also updates PT note copy to "You can only select public transport stops
on KRAIL App."

feat: stop label save/manage workflow

- Add `+ Save` trailing chip on each stop row → opens SaveStopAsLabelSheet
  for picking which label to assign the stop to (or create a new one).
- Long-press any pill → ManageLabelSheet with Clear stop / Delete label.
- New ClearLabelStop and DeleteLabel events wired through ViewModel.
- Stop label icons (home, work, gym, cafe, beach, hospital) drive both
  the suggestion chips in AddLabelBottomSheet and the pill icons in the
  shortcuts row.
- Fix AddLabelBottomSheet preview animation: background pill is now
  static; only the icon+text inside slide/fade as the user types.
- Inline label pills directly in SearchStopScreen and remove the now
  orphaned StopLabelPillRow.kt.

fix: instant pill updates, star toggle, keyboard dismiss, airport icon

- Optimistic state updates for AssignLabelStop / CreateLabel /
  ClearLabelStop / DeleteLabel so the pill row reflects changes
  immediately, without waiting for the DB write to round-trip.
- Move observeStopLabels() to init {} so the DB observer runs for the
  full VM lifetime instead of restarting on every UI subscription.
- Replace the `+ Save` chip on stop rows with a star icon: filled when
  the stop is saved as any label, outlined otherwise. Reuses the same
  ic_star / ic_star_filled icons used for saved trips.
- Dismiss keyboard and free focus when tapping a set label pill so it
  matches the regular stop-select navigation path.
- Show the airport icon next to transport mode chips on stop rows whose
  name contains "airport" (sized to match TransportModeIconSize.Small,
  tinted with KrailTheme.colors.label).

feat: drag-to-reorder pills, label conflict warning, dedupe create

- Long-press a pill enters edit mode; pills wiggle (rotation animation
  with per-pill timing offset) and become drag-reorderable. Drag past
  60% of a pill's width swaps with the neighbour and renumbers all sort
  orders via MoveLabelToIndex.
- Tap "Done" pill (replaces "+ Add" while editing) to exit. Tap any
  pill in edit mode to open ManageLabelSheet for clear/delete.
- New LabelConflictSheet warns when assigning a stop already saved to
  another label. Confirming clears the previous label and assigns the
  new one; cancelling does nothing.
- Dedupe CreateLabel optimistic update: if a label with the same name
  already exists (case-insensitive), skip both the state append and the
  DB write. Previously a duplicate name in optimistic state caused the
  LazyRow keyed by label.label to crash with "Key already used".
- Use a stable item key for the trailing "+ New label" chip in
  SaveStopAsLabelSheet so a user-created label named "new" cannot
  collide with it.

feat: pill row UX overhaul — reorderable lib, banner, conflict warnings

Drag-and-drop reorder
- Replace the brittle hand-rolled drag implementation with sh.calvin.reorderable.
  Drag-from-end-to-start, alignment after release, ripple containment, and the
  "stuck mid-drag" issue are all handled by the library.
- Long-press a pill (outside edit mode) → enters edit mode. Pills wiggle.
- In edit mode: pills accept drag immediately via draggableHandle; tap is a
  no-op so the manage sheet can't pop mid-drag. ManageLabelSheet still
  reachable from non-edit mode via long-press.
- Modifier order fixed (graphicsLayer → clip → clickable) so the long-press
  ripple is now contained inside the rounded pill, not a rectangle.

Star toggle
- Filled star on a stop row now fires ClearLabelStop directly — no sheet —
  for the matching label. Outlined star opens SaveStopAsLabelSheet as before.

Assigning-mode banner
- When assigningLabel is set (either by tapping an unset pill or after
  creating a new label), an inline banner appears under the pill row:
  "Tap the ⭐ next to a stop to save it as <name>" with a Cancel action.
- Creating a new label auto-enters assigning mode for that label so the user
  knows what to do next.

Conflict warnings (both directions)
- LabelConflictSheet generalised to title + message + confirmLabel.
- Stop-side conflict: same stop already saved as a different label →
  "Already saved … Move to Y?".
- Label-side conflict: chosen label already has a different stop →
  "Already in use … Replace with Y?".

AddLabelBottomSheet redesign
- Title is now "Add a new label". Sections stacked: Title → (StopChip if any)
  → Preview heading + pill → Suggestions heading + chips → Name heading +
  TextField → Save. Consistent vertical rhythm via Arrangement.spacedBy.
- Removed the AnimatedContent on the preview pill; renders statically.
- Suggestions filter out names that already exist (case-insensitive,
  emoji-stripped) so Gym disappears once Gym is added.
- Inline duplicate error under the TextField + Save disabled when the typed
  name (after normalising) already exists.

Name normalisation
- New normaliseLabelName() strips emoji and exotic characters and collapses
  whitespace. Used on save and for dedupe so `🏠 Home`, `Home`, and `home`
  all resolve to the same canonical name.

SearchStopScreen polish
- Recent header uses pageHorizontalPadding (matches stop rows below); the
  clear-all "✕" sits in a 36 dp tap target, aligning with the star buttons.
- Recent text demoted to titleMedium/softLabel — less visually heavy.
- Pill row contentPadding tightened (spacingS vertical) so it doesn't
  hover with extra whitespace.
- Stop-row dividers now padded by pageHorizontalPadding instead of going
  edge-to-edge.

State previews
- Added @PreviewScreen entries: fresh install, labels-only no recents,
  recents-only no labels, all labels + recents, and assigning mode. These
  feed the existing snapshot pipeline.

feat: long-press to drag, inline ✕ delete, set-first sort, Clear all text

Long-press → drag in one motion
- Replace draggableHandle with longPressDraggableHandle (enabled only while
  editing). Long-press a pill while in edit mode triggers drag immediately —
  no separate "press, release, press again" step.
- combinedClickable.onLongClick (outside edit mode) is what enters edit
  mode in the first place. The lambda is always non-null so combinedClickable
  consumes the long-press, preventing onClick from firing on long-hold release.

Tap is always "select"
- Tap a set pill → select as From/To. Tap an unset pill → enter assigning
  mode. Both work whether or not we're in edit mode. No bottom sheet on tap.

Inline ✕ delete (replaces ManageLabelSheet)
- In edit mode, each pill shows a small ✕ chip in the top-right corner that
  fires DeleteLabel directly. ManageLabelSheet is removed; clearing a stop is
  done via the filled star on stop rows.

Set labels sorted first
- selectAllStopLabels SQL now ORDERs BY (stop_id IS NULL) ASC first, so
  labels with an assigned stop appear at the start of the row, with unset
  labels grouped after. Sort order within each group is preserved, so
  drag-to-reorder still works inside the set/unset groups.

Recent header
- "✕" icon button replaced with a "Clear all" text button, matching the
  surrounding label/colour scheme.

feat: protect Home, save sheet outlined unset chips, awaitEachGesture, savers

Home is non-deletable
- Add StopLabel.isProtected (true for "Home"). The pill row hides the ✕
  delete chip on protected pills, and DeleteLabel in the ViewModel early-
  returns for protected names. Stops the previous bug where deleting both
  defaults caused observeStopLabels' empty-rows branch to reseed Home and
  Work back in.

Save sheet chip styling
- LabelChoiceChip is now outlined (theme-color border) when the label is
  unset and solid only when it already has a stop. Tapping an outlined chip
  attaches this stop to that empty label as before — the visual just
  matches the actual state now.

Custom gesture detector replaces combinedClickable
- Drop combinedClickable from the pill modifier. It was racing with
  longPressDraggableHandle for the same long-press event, which is why
  drag never engaged.
- Replace with pointerInput { awaitEachGesture { … } } that explicitly
  branches on (released within timeout = tap, timeout reached = long-press,
  cancellation inside timeout = let drag handle take over). The drag
  handle now reliably gets long-press + drag.

Saveable UI orchestration state
- Add mapSaver-based StopLabelSaver, StopItemSaver and LabelConflictSaver
  so the four bits of UI state (assigningLabel, stopBeingSaved,
  pendingConflict, editingLabels, showAddLabelSheet) survive rotation,
  dark-mode toggle, locale change and process death. Was: rotating
  mid-assign or mid-conflict dropped the sheet.

fix: nullable Savers for rememberSaveable state

Compile error: rememberSaveable(stateSaver = ...) infers MutableState<T>
from the Saver's Original type. mapSaver requires Original : Any, so
mapSaver-built Savers couldn't back nullable MutableState<StopLabel?> etc.

Replace with hand-written Saver(save, restore) blocks where save returns
null when the value is null (the framework treats null as "nothing to
persist"), and restore is only called when there is a saved map.

feat: SearchStopScreen polish — Done button, sticky search header, tokens

Done button distinct from pills
- Black/white onSurface background (not theme color), so Done reads as an
  action button rather than yet another label pill. Extra leading gap
  before it (spacingL) so the visual separation is obvious.

Sticky search-status header
- Always-rendered SearchingDotsHeader at the top of the LazyColumn, in
  both Recent and Results states. Pill row stays visually anchored when
  the user starts typing — the dots area no longer pushes everything
  down. Removed the bottom padding under the SearchTopBar TextField for
  the same reason.

Recent header colour + Add pill
- Recent label and "Clear all" now use KrailColors.label (was softLabel)
  for stronger contrast.
- "+ Add" pill border + text now use label colour (was outlineSubtle /
  softLabel), making the affordance more visible.

Layout cleanup
- Removed the divider below the pill row — it was crowding spacing.
- Reduced pill row top padding from spacingM to spacingS now that the
  ✕ delete chip overhang is smaller.

Realistic previews
- Replaced unrealistic "no labels" previews with focused scenarios that
  always include the seeded Home/Work pills + recents:
  FreshInstall, Recent, RichLabels, AssigningMode, SearchLoading,
  SearchResults, NoMatch, Error, Map. Shared previewRecentStops /
  previewLabelsTypical / previewLabelsRich fixtures cut duplicate
  boilerplate.

Tokens everywhere
- Replaced every hardcoded .dp value in SearchStopScreen.kt with
  KrailTheme.dimensions tokens (pageHorizontalPadding / spacingS /
  spacingL / spacingXL / spacingXXL / spacingXXXXL / spacingNone).
  Dropped the now-unused androidx.compose.ui.unit.dp import.

feat: previews use PreviewTheme(themeStyle), label colours, vary themes

- All SearchStopScreen previews now drive theme colour via
  PreviewTheme(themeStyle = …) instead of building their own
  CompositionLocalProvider + LocalThemeColor block. Drop the unused
  CompositionLocalProvider and NswTransportMode imports.
- Spread theme styles across previews so snapshots cover the palette:
  FreshInstall = Train, Recent = Bus, RichLabels = Metro,
  AssigningMode = BarbiePink, SearchLoading = Ferry,
  SearchResults = Train, NoMatch = Metro, Error/Map = PurpleDrip.
- Set isMapsAvailable + mapUiState in the relevant idle/loading/results
  previews so the "Select on map" affordance now appears in coverage.
- UnsetLabelPill border + content colour switch from outlineSubtle /
  softLabel to KrailColors.label so unset pills stay readable on
  every theme background.
- AssigningModeBanner copy now uses label (was softLabel).
- PublicTransportNote text uses label (was softLabel) and horizontal
  padding switched to pageHorizontalPadding for consistency with the
  rest of the screen.

feat: floating dots pill, toggleable assigning banner

Compact searching… indicator
- SearchingDotsHeader is now a small rounded pill (onSurface background,
  themeColor dots) that floats on the z-axis above the list, instead of
  a full-width 32 dp row claiming a chunk of vertical space at the top.
- Wrapped each LazyColumn in a Box so the pill can sit in TopCenter and
  fade/scale in/out via AnimatedVisibility.
- Removed the dedicated "search-status" item that used to anchor the
  layout — the LazyColumn no longer reserves space for it.

Toggleable assigning banner
- Tap an unset pill: enter assigning mode for it.
- Tap the same pill again: exit assigning mode (or tap a different
  unset pill to switch). No more separate Cancel button.
- Banner item is always present in the LazyColumn; its content is
  conditional. Wrapped in Modifier.animateContentSize so the height
  smoothly expands/collapses on toggle.

Banner copy uses label colour (already; restated in the simpler
single-Text composable now that Cancel is gone).

Dropped the now-unused onCancelAssigning callback chain from
SearchStopScreen / SinglePane / DualPane / SearchStopListContent.

feat: shared LoadingDotsPill in taj, contextual banner, SelectOnMap align

Reusable taj components
- Move AnimatedDots into taj.components and add a new
  LoadingDotsPill wrapper alongside it. The pill bakes in the
  visibility-with-min-visible-time state machine, the rounded
  onSurface chip styling and a fixed dp canvas size so the dots
  actually render. Other features (DepartureBoard, TimeTable,
  TrackTrip, DepartureBoardStopSection) repointed at the new path.
- SearchingDotsHeader is now a thin alias forwarding to
  LoadingDotsPill so existing call sites keep working.

Contextual hint banner
- New pure pillRowBannerText(editing, assigningLabel, stopLabels)
  picks the right copy:
  • editing = true → "Drag a pill to reorder. Tap ✕ to delete."
  • assigningLabel set & still unset → "Tap the ⭐ next to a stop to
    save it as <label>"
  • everything else → null (no banner).
  Pure + injectable, easy to unit-test in isolation.
- Renamed AssigningModeBanner → PillRowInfoBanner; takes a single
  text param now.
- LaunchedEffect clears assigningLabel automatically when the label
  becomes set (e.g. via the save sheet), so the hint collapses on
  its own without manual cancel.

SelectOnMap alignment
- Padding switched from spacingXL/spacingXXL to the same vertical /
  horizontal values stop rows use (spacingM / pageHorizontalPadding),
  with arrangement spacing matching too. The icon and "Select on map"
  text now line up flush with the stop names below.

test: extract pure rule helpers, add unit tests, doc invariants

Quick-win unit-test pass before the bigger Compose UI test branch.

Pure helpers
- New SearchStopRules.kt collects screen orchestration logic that's a
  pure function of state: pillRowBannerText, savedStopIds and
  conflictForAssign (with an internal AssignConflict sealed interface).
- The previous inline pillRowBannerText in SearchStopScreen.kt is gone;
  the screen now imports the rules-file version.
- Anything we add later that's "given (state) → (decision)" should land
  here so screen code stays thin and tests stay easy.

Tests
- SearchStopRulesTest covers all banner branches (idle / editing /
  assigning unset / assigning satisfied / unrelated label change),
  savedStopIds, and every conflictForAssign outcome (no-conflict,
  same-stop reassignment, stop-side, label-side, both-with-precedence).
- LabelNameNormalizerTest covers normaliseLabelName + labelNamesMatch
  (whitespace, embedded emoji, punctuation drop, punctuation we keep,
  empty-input edge cases, case-insensitive match).

Documentation
- feature/trip-planner/ui/SEARCH_STOP_UX.md captures every UX invariant
  on the screen — gestures, contextual banner, conflict logic, sort
  order, state persistence — so the rules can't quietly drift.
  Includes a "TODO / future work" list of Compose UI tests + snapshot
  states scheduled for a follow-up branch.
- CLAUDE.md gets a "Per-feature UX rule docs" pointer so future agent
  runs read the doc first.

feat: AddLabel description, reorder copy update, recent header padding

- AddLabelBottomSheet now shows a bodySmall description below the title
  when no stop is attached, explaining what labels are and how to attach
  one ("Labels are quick shortcuts to your favourite stops like Home,
  Work, or Gym, anywhere you ride to often. Pick a name now, then tap
  the ⭐ next to a stop in your search results to attach it.").
- Edit-mode banner copy switched to "Long press and then drag the pill
  to reorder and select Done to save." The ✕ button is self-evident so
  the reorder hint stays focused on the gesture, and the matching
  SearchStopRulesTest assertion is updated.
- "Recent" header gets extra top breathing room (spacingXL) so it
  doesn't crowd whatever sits above it.
- Select-on-map row is now flanked top + bottom by Divider rules inset
  to pageHorizontalPadding (extracted as DividerInset helper) so it
  reads as its own visual section between the pill row and the recent
  list.

chore: drop redundant else in JourneyCard deviation when

showDeviation is only true when departureDeviation is Late or Early,
so smart-casting makes the when exhaustive without an else branch.
Killing the warning surfaced during the last compileAndroidMain run.

Revert "chore: drop redundant else in JourneyCard deviation when"

This reverts commit f8ea994dbde5821d305964f05920ca0f5b1b834b.

feat: dedupe lazy items, hide pill row on empty, save-to-recents, info toggle

LazyListScope helpers
- pillRowSection (label shortcuts + assigning banner), selectOnMapItem,
  publicTransportNoteItem extracted into LazyListScope extensions.
  SearchStopListContent's Recent and Results branches no longer copy-
  paste the same item blocks; both call the helpers.

Hide pill row when list has nothing to act on
- New shouldShowPillRow(listState, recentStops) rule:
  • Recent: visible only when there's at least one recent stop.
  • Results: visible only when results are non-empty (so loading,
    NoMatch and Error all hide the row).
- Reasoning: tapping an unset Home pill enters assigning mode, which
  expects star buttons below. Showing pills against an empty canvas
  would let users tap into a dead-end. Six new unit tests in
  SearchStopRulesTest cover every state.

Saving a stop also pins it to Recents
- AssignLabelStop handler now also calls
  sandook.insertOrReplaceRecentSearchStop(stopId) and refetches
  recentStops, so a stop attached via the save sheet shows up in
  Recents next time without forcing the user to re-search.

Recent header restyle
- "Recent" promoted to titleLarge with FontWeight.Normal (slightly
  bigger heading, not bold).
- "Clear all" demoted to bodyMedium so it reads as a tertiary action,
  not a button.

Info-toggle in Add a new label sheet
- Replaces the always-visible help paragraph with a small ic_info
  icon next to the title. Tapping it toggles the explanation in/out
  with animateContentSize for a smooth height transition.
- Default is collapsed; rememberSaveable so the user's last choice
  survives rotation.

Docs
- SEARCH_STOP_UX.md now documents the pill-row visibility rule and
  the auto-pin-to-recents behaviour. Compose UI test coverage for
  these rules is queued in the TODO section for the follow-up branch.

fix: keep stop context across "+ New label" and conflict-cancel flows

Two related bugs in the save-as-label flow where the stop the user
originally tapped a star on would be lost:

"+ New label" auto-attach
- Tapping "+ New label" inside SaveStopAsLabelSheet was clearing
  stopBeingSaved before opening AddLabelBottomSheet, so creating a
  brand-new label like "Cinema" left the stop unattached and forced
  the user to tap the star again.
- Add pendingStopForNewLabel state (saveable). Set when the user
  enters the new-label flow from SaveStopAsLabelSheet; consumed in
  AddLabelBottomSheet.onSave to fire AssignLabelStop alongside
  CreateLabel. The "+ Add" pill flow (no pendingStop) keeps its
  existing behaviour of dropping into assigning mode.
- AddLabelBottomSheet.onDismiss / blocked-save now hand the stop
  back to SaveStopAsLabelSheet via stopBeingSaved = pending, so the
  user can pick a different label instead of being stranded.

Conflict cancel
- LabelConflictSheet's onCancel was just clearing pendingConflict
  with no path back. Now it also restores stopBeingSaved =
  conflict.stop, so cancelling reopens SaveStopAsLabelSheet for the
  user to try a different label.

fix: scale-up assigning pill, single star tint, sealed conflict.stop

Pill assignment uses scale, not colour
- Tinting an unset pill themeColor when it's the assigning target
  blended into the screen's themed background gradient. Switched the
  feedback to a scale animation (1.10× via animateFloatAsState) on
  the parent Box, matching the departure-board chip's growing-on-tap
  language.
- UnsetLabelPill border + content stay KrailColors.label regardless
  of assigning state. The isAssigning param is kept on the public
  signature for now (suppressed) since callers still pass it; will
  remove with the next refactor pass.

Star tint
- SaveAsLabelStar now uses KrailColors.label for both the filled and
  outlined variants. Only the icon silhouette changes between states,
  so the colour stays consistent with surrounding label text.

Sealed conflict.stop
- LabelConflict promotes target + stop to the sealed interface, so
  callers can read conflict.stop without a smart-cast. Fixes the
  unresolved-reference compile error introduced by the conflict-cancel
  restoration code in the previous commit.

feat: rotate star 360° on save state transition

When a stop's saved state actually flips (because the user picked a
label in SaveStopAsLabelSheet, or unsaved by tapping a filled star),
spin the star icon a full 360° using an Animatable + tween — same
animation the intro screen uses for save-trip.

The rotation only fires on transitions: a `hasInitialised` flag skips
the first composition, so the star doesn't spin every time the row
scrolls into view. Subsequent isSaved toggles trigger the animation.

This decouples the visual feedback from the tap (which only opens the
sheet) and ties it to the moment the assignment actually completes,
which is what the user sees.

chore: drop dividers around Select-on-map row

The dividers above and below the Select-on-map row were redundant given
the existing vertical rhythm from the pill row above and the stop-row
spacing below. Remove both, plus the now-orphaned DividerInset helper.

fix: star rotation, edit-mode tap suppression, Done at row start

Star rotation (and the "save didn't work" bug behind it)
- StopSearchListItem was rendering SaveAsLabelStar from two different
  source positions inside a `when {}`. Compose keys composable state
  by source position, so flipping isSaved was destroying the previous
  composition and creating a fresh one — Animatable + hasInitialised
  reset every flip, so the rotation never fired AND the user perceived
  the save flow as broken because the star didn't visibly update.
- Now compute the click handler conditionally and call SaveAsLabelStar
  from a single source position. State survives across isSaved changes;
  rotation animates as expected.
- Took the opportunity to swap the boolean param for a StarState enum
  (Saved / Unsaved) per the request — clearer call sites and a single
  key for LaunchedEffect to drive the transition off.

Edit-mode taps suppressed
- The pointerInput's awaitEachGesture now no-ops on tap when
  editingLabels is true. Only drag (via longPressDraggableHandle) and
  the ✕ delete chip respond. Accidental brushes against a pill while
  reordering no longer navigate away or enter assigning mode.

Done pill moved to row start
- The Done pill now occupies a leading "leading-done" item before the
  pills (with end padding so it doesn't crowd the first pill). The
  trailing "+ Add" slot is hidden while editing. Done is always
  reachable without scrolling past the user's pills.

fix: directional star rotation, match intro-screen timing

The save-as-label star now spins clockwise on save and
counter-clockwise on remove, so the gesture reads like "winding it in"
vs "letting it go" rather than two identical-looking 360s. Duration
bumped from 400 ms to 500 ms so it matches the intro screen's
save-trip animation — the two surfaces now feel like the same
affordance.

fix: AnimatedDots density-aware, LoadingDotsPill canvas fits bounce

The dots were always rendering off-centre inside LoadingDotsPill — at
high densities the canvas was too short for the dots to bounce without
clipping at the top, and at low densities the dots looked oversized
because the geometry was specified in raw pixels instead of dp.

AnimatedDots
- dotRadius (4 dp default), dotSpacing (14 dp), bounceHeight (10 dp)
  are now Dp parameters that get converted to pixels at draw time per
  current density. Animation frames are unit-less fractions in [-1, 0]
  scaled to bounceHeight, so the bounce reads the same on every phone
  and changing bounceHeight just makes the dots bob higher/lower.

LoadingDotsPill
- Canvas bumped to 44 × 32 dp (was 48 × 24 dp). The new height fits
  2 × (radius + bounceHeight) = 28 dp + a tiny margin so the dots
  never kiss the pill border at peak bounce. Width still gives a
  couple of dp of horizontal breathing room either side of the dots.
- Pill horizontal padding tightened from spacingL → spacingM so the
  total pill stays compact now that the canvas is taller.

chore: extract magic numbers + suppress LongMethod with rationale

Pre-PR polish so detekt runs clean and reviewers don't have to ask
about constants pulled out of thin air.

Magic-number extraction
- Pill scale: PILL_RESTING_SCALE / PILL_DRAG_SCALE /
  PILL_ASSIGNING_SCALE in SearchStopScreen.
- Wiggle animation: WIGGLE_ANGLE_DEGREES + four duration / offset
  variance constants describing why each pill bobs at slightly
  different timing.
- Star rotation: STAR_ROTATION_DEGREES / STAR_ROTATION_DURATION_MS /
  SAVE_ROTATION_DIRECTION / UNSAVE_ROTATION_DIRECTION + the star tap
  target / inner padding dp values, in StopSearchListItem.
- Loading pill enter/exit scale: ENTER_EXIT_SCALE in LoadingDotsPill.

LongMethod suppressions with explicit rationale
- SearchStopScreen, LabelShortcutsRow, AddLabelBottomSheet each now
  carry @Suppress("LongMethod") with a comment explaining why
  splitting hurts more than it helps (state plumbing / gesture
  coupling / single-flow sheet UX). Detekt flags 60-line methods by
  default; suppressing here keeps the warning surface clean for the
  rules detekt would actually catch.

docs: align SEARCH_STOP_UX banner copy with implementation

The editing-mode banner was changed in 5700b51 to the long-press wording
but the doc table still showed the old "drag a pill / tap ✕" copy. Sync
the doc so reviewers don't flag the discrepancy.

test: implement StopLabels methods on FakeSandook

The Sandook interface gained observeStopLabels / upsertStopLabel /
updateStopLabelStop / deleteStopLabel / clearStopLabels in this branch
but FakeSandook still inherited them as abstract — every test using
FakeSandook fails to compile. Add an in-memory backing flow that mirrors
RealSandook's sort order so VM tests can drive label flows.

test: pass preferences + sandook to SearchStopViewModel constructor

The VM gained two new constructor parameters in this branch (preferences
for the maps-options first-time flag, sandook for label persistence).
Existing tests were still passing the old 5-arg form and failing to
compile. Wire FakeSandookPreferences + FakeSandook through.

docs: track resolved test gaps in SEARCH_STOP_UX

Promote the Compose UI tests + new ViewModel handler tests from "TODO"
to "Covered", and pin the small set of outstanding cases (gesture-heavy
or sheet-animation-bound) with explicit rationale so future readers
don't re-attempt them blindly.


ksharma-xyz:
* **#1523** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1523?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1522** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1522?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1521** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1521?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1521?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1524, opened 2026-04-29
ci: run host tests in code-quality job after detekt

ksharma-xyz:
## Summary
- Adds \`./gradlew testAndroidHostTest --continue\` as a step inside \`code-quality.yml\`, right after detekt.
- Failing tests now gate \`build-android-debug\`, \`build-android-release\` and \`build-ios\` exactly the way detekt failures do today (those jobs all \`needs: code-quality\` in \`build.yml\`).
- Compose UI tests (Robolectric, in \`androidUnitTest\`) run alongside pure-function and ViewModel tests via KMP's \`withHostTest {}\`-registered task.
- \`--continue\` so every failing module is reported in one run.
- Test reports are uploaded as a separate artifact (\`test-reports-<run-id>\`, retention 7 days).
- The test step uses \`if: always() && steps.detekt.conclusion != 'cancelled'\` so a red detekt run still surfaces test failures in the same CI pass — saves a re-run cycle.

## Why now
Today CI gates on detekt + 3 platform builds, but never executes a single test. \`feat/home-stop-labels\` added ~50 new tests across rules, ViewModel handlers, and Compose UI; main has TimeTableViewModel + DepartureBoard tests already. None of them block a regression in CI.

## Trade-off
Single combined \`code-quality\` job (per your suggestion) vs. a parallel \`tests\` job. Combined is simpler — one workflow, one cache warm-up, one set of secrets — at the cost of a slower single job. If runtime becomes a problem we can split later.

## Test plan
- [ ] Push to a draft PR, watch the \`code-quality / detekt\` check run both detekt + tests.
- [ ] Confirm test artifact uploads on red and green runs.
- [ ] Confirm \`build-android-*\` / \`build-ios\` jobs still gate on the combined job (no config change to \`build.yml\`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1528, opened 2026-04-30
refactor: extract TripPoller from TrackTripViewModel

ksharma-xyz:
The ViewModel owned three while(isActive) loops (poll job, GTFS-RT
live-positions job, clock job) plus the smart-delay anchor and
fetchAndUpdate IO. They were untestable in isolation: the only seam
into them was viewModelScope, which has no public cancellation hook,
so VM tests had to wrap runTest in a helper that cancelled
viewModelScope in a finally block to stop the loops before runTest's
end-of-body advanceUntilIdle drained them forever.

Moves all of that into a new TripPoller class that takes its
CoroutineScope in the constructor:
- TrackTripViewModel constructs TripPoller(viewModelScope, …) and
  delegates lifecycle (startPolling, startLivePositionPolling,
  transitionToArrivedAndFinished, isTripExpired, reset)
- TripPoller owns clock, liveOverlay, stopCoordinates, countdownDisplay
  StateFlows; the VM just re-exposes them so the public API of
  TrackTripViewModel is unchanged (TrackTripScreen and the Koin module
  don't move)
- Public/internal split: TripPoller is internal — it's an
  implementation detail of the track:ui module

The new TripPollerTest exercises the polling logic directly via
runTest { TripPoller(backgroundScope, …) }. backgroundScope is auto-
cancelled when the test ends so the loops stop without manual cleanup.
TrackTripViewModelTest still uses the runTrackingTest wrapper because
it constructs the VM (which still binds the poller to viewModelScope);
the docstring on the wrapper now points at TripPollerTest as the
cleaner alternative.

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

### DIFF
--- a/feature/track/ui/src/commonMain/kotlin/xyz/ksharma/krail/feature/track/ui/TrackTripViewModel.kt
+++ b/feature/track/ui/src/commonMain/kotlin/xyz/ksharma/krail/feature/track/ui/TrackTripViewModel.kt
@@ -2,7 +2,6 @@
     "LongMethod",
     "TooManyFunctions",
     "CyclomaticComplexMethod",
-    "LoopWithTooManyJumpStatements",
     "ForbiddenComment",
 )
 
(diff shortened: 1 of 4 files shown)


ksharma-xyz:
* **#1528** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1528?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1528?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1524** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1524?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #1556, opened 2026-05-03
feat(analytics): track stop label create, assign, and remove events

ksharma-xyz:
## Summary
Adds three high-signal analytics events covering the stop labels lifecycle (create → assign → remove) so BigQuery can answer "how are users actually using stop labels?" from a single dataset, plus ViewModel tests that lock down the wire-up.

### Events
1. **`stop_label_created`** — adoption signal.
   Properties: `labelName`, `emoji`, `totalLabelsCountAfter` (so casual vs heavy users can be split).
2. **`stop_label_stop_assigned`** — the core "label is being used" signal, fired when a stop is pinned to a label.
   Properties: `labelName`, `stopId`, `stopName`, `isReassignment` (overwriting an existing pin vs filling an empty slot), `isProtectedLabel` (true for Home).
3. **`stop_label_removed`** — unified cleanup event.
   Properties: `labelName`, `action` (`"delete"` or `"clear"`), `hadStop` (was a stop attached at the moment of removal).
   Note: protected Home-delete attempts are silently no-op'd by the handler, so this event does not fire for them — keeps the dataset honest from defensive UI calls.

### Why these three
- Cover the full label lifecycle (create → fill → remove), so funnel analysis works without joining multiple event streams.
- Carry enough metadata to answer the most likely product questions (which labels are popular, which stops people pin most, are users replacing pins or filling slots, are users abandoning labels they never filled).
- All three already have ViewModel-native handler entry points — no UI-layer instrumentation needed.

## Where
- `core/analytics/.../AnalyticsEvent.kt` — three new sealed-class entries in a new `StopLabels` region. New `PROP_LABEL_NAME` private const to satisfy the existing `StringLiteralDuplication` rule.
- `feature/trip-planner/ui/.../searchstop/SearchStopViewModel.kt` — `analytics.track(...)` calls wired into `CreateLabel`, `AssignLabelStop`, `ClearLabelStop`, and `DeleteLabel` handlers. Pre-mutation snapshots used so `isReassignment` and `hadStop` reflect state *before* the optimistic update.
- `core/test/.../SearchStopViewModelLabelHandlersTest.kt` — eight new tests in a dedicated `Analytics` region covering: created event payload, no-emit on duplicate / blank name, first-pin vs reassignment, clear with `hadStop=true`, delete with `hadStop=false`, and silent no-op on protected Home delete.

## Test plan
- [x] `./gradlew :core:test:testAndroidHostTest --tests 'xyz.ksharma.core.test.viewmodels.SearchStopViewModelLabelHandlersTest'` — all green.
- [x] `./gradlew :core:analytics:detekt :feature:trip-planner:ui:detekt :core:test:detekt --continue` — clean.
- [ ] Manual sanity check on a debug build: create a label, pin a stop to it, clear it, delete a custom label, attempt to delete Home → confirm the four expected events appear in the analytics console (and the Home-delete attempt does not).

### DIFF
--- a/core/analytics/src/commonMain/kotlin/xyz/ksharma/krail/core/analytics/event/AnalyticsEvent.kt
+++ b/core/analytics/src/commonMain/kotlin/xyz/ksharma/krail/core/analytics/event/AnalyticsEvent.kt
@@ -8,6 +8,7 @@ private const val PROP_STOP_ID = "stopId"
 private const val PROP_SOURCE = "source"
 private const val PROP_EXPAND = "expand"
 private const val PROP_STOP_NAME = "stopName"
+private const val PROP_LABEL_NAME = "labelName"
 
 sealed class AnalyticsEvent(val name: String, val properties: Map<String, Any>? = null) {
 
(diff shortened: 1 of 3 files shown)


ksharma-xyz:
* **#1556** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1556?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1556?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1562, opened 2026-05-03
feat(analytics): track saved-trip + label reorder, plus timetable stop-header click

ksharma-xyz:
## Summary
Adds three high-signal analytics events that close gaps in the saved-trips and timetable dashboards. Same disciplined approach as #1556 — every event carries enough context to answer specific product questions without joining multiple streams, and each path has a ViewModel test.

### Events

**1. `saved_trip_card_reordered`** — wired into `SavedTripsViewModel.onMoveSavedTrip`.
- `fromStopId`, `toStopId` — anchors the moved card to a specific saved trip (matches the `SavedTripCardClickEvent` pattern, so dashboards can join).
- `previousIndex`, `newIndex` — direction + delta of the move.
- `totalCount` — list size at reorder time, splits casual users (2 cards) from heavy users (8+).
- Handler also tightened: source == target now hard-returns before any state write or analytics fire, so drop-in-place gestures don't inflate counts.

**2. `stop_label_reordered`** — wired into the existing `SearchStopUiEvent.MoveLabelToIndex` handler in `SearchStopViewModel`.
- `labelName`, `previousIndex`, `newIndex`, `totalCount` — same shape as event 1.
- `isProtectedLabel` — true when the moved label is the protected Home. Lets the dashboard separate "user reorders Home" (which is non-deletable but draggable) from "user reorders custom labels".

**3. `timetable_stop_header_click`** — the gesture that opens the stop-details bottom sheet from the timetable sticky header.
- New analytics-only `TimeTableUiEvent.OriginDestinationStopHeaderClicked(stopId, stopName, isOrigin)` fired by the UI; the VM enriches with the trip context.
- Payload: `stopId`, `stopName`, `isOrigin` (origin vs destination), `tripFromStopId`, `tripToStopId`.
- `isOrigin` is the highest-signal field: "are users more curious about where they're leaving from, or where they're going to?". The trip pair anchors back to `PlanTripClickEvent` / `LoadTimeTableClickEvent` so "of users who plan a trip, what % drill into stop details" is one join away.

## Why these three (and not more)

- All three previously had **zero** instrumentation despite being core gestures.
- Each event answers a distinct product question; payloads were sized to make that question answerable from BigQuery without joining sister streams.
- Skipped: per-side stop tap on saved-trip cards (already covered by `SavedTripCardClickEvent` — the card is one click target, not two), stop-info dwell time (the existing `DepartureBoardScreenViewEvent` already fires inside the sheet), drag duration / distance / cancelled-drag attempts (low signal, hard to test).

## Test plan
- [x] `./gradlew :core:test:testAndroidHostTest --tests SavedTripViewModelsTest --tests SearchStopViewModelLabelHandlersTest --tests TimeTableViewModelTest --continue` — green.
- [x] Detekt clean on `:core:analytics`, `:feature:trip-planner:state`, `:feature:trip-planner:ui`, `:core:test`. Cleaned up pre-existing `"fromStopId"` / `"toStopId"` raw-string duplications by promoting them to shared `PROP_FROM_STOP_ID` / `PROP_TO_STOP_ID` constants now that more events share the keys.
- [ ] Manual sanity: enter saved-trips edit mode → drag a card → expect `saved_trip_card_reordered`. Long-press a stop label and drag it → expect `stop_label_reordered`. Plan a trip → on the timetable, tap origin and destination labels in the sticky header → expect two `timetable_stop_header_click` events with `isOrigin` flipped between them.

### DIFF
--- a/core/analytics/src/commonMain/kotlin/xyz/ksharma/krail/core/analytics/event/AnalyticsEvent.kt
+++ b/core/analytics/src/commonMain/kotlin/xyz/ksharma/krail/core/analytics/event/AnalyticsEvent.kt
@@ -9,6 +9,11 @@ private const val PROP_SOURCE = "source"
 private const val PROP_EXPAND = "expand"
 private const val PROP_STOP_NAME = "stopName"
 private const val PROP_LABEL_NAME = "labelName"
+private const val PROP_FROM_STOP_ID = "fromStopId"
+private const val PROP_TO_STOP_ID = "toStopId"
+private const val PROP_PREVIOUS_INDEX = "previousIndex"
+private const val PROP_NEW_INDEX = "newIndex"
+private const val PROP_TOTAL_COUNT = "totalCount"
 
 sealed class AnalyticsEvent(val name: String, val properties: Map<String, Any>? = null) {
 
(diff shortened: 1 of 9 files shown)


ksharma-xyz:
* **#1562** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1562?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1562?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1594, opened 2026-05-16
fix(navigation): prevent transient NoEntriesUI flash

ksharma-xyz:
fix(navigation): prevent transient NoEntriesUI flash

Two bugs caused NoEntriesUI to appear randomly:

1. KrailNavHost called toEntries() twice — once for the guard check and
   again inside NavDisplay. Each call is a separate @Composable call-site,
   so Compose gives them independent rememberDecoratedNavEntries slots.
   The guard checked instance A while NavDisplay received instance B,
   creating a window where they could diverge. Fixed by reusing the single
   computed `entries` val in NavDisplay.

2. NavigationState.resetRoot() called clear() then add() as two separate
   SnapshotStateList mutations. Between those two writes Compose could
   observe an empty back stack and trigger a recomposition that hit the
   else branch. Same risk existed in replaceCurrent(). Fixed by wrapping
   both pairs in Snapshot.withMutableSnapshot{} so the mutations are
   applied atomically and no intermediate empty state is ever visible.

analytics(navigation): track no_entries_detected for fallback UI monitoring

Keeps the NoEntriesUI fallback in place as a safety net, but now fires a
no_entries_detected event (with topLevelRoute) whenever entries go empty so
we can monitor Firebase to confirm the root-cause fixes (duplicate toEntries
call + non-atomic resetRoot mutations) eliminated the condition in production.
If the event stays silent after a release, the fallback and this event can be
removed together.

### DIFF
--- a/composeApp/src/commonMain/kotlin/xyz/ksharma/krail/KrailNavHost.kt
+++ b/composeApp/src/commonMain/kotlin/xyz/ksharma/krail/KrailNavHost.kt
@@ -23,6 +23,8 @@ import androidx.compose.ui.unit.dp
 import androidx.navigation3.runtime.NavKey
 import androidx.navigation3.ui.NavDisplay
 import org.koin.compose.koinInject
+import xyz.ksharma.krail.core.analytics.Analytics
+import xyz.ksharma.krail.core.analytics.event.AnalyticsEvent
 import xyz.ksharma.krail.core.deeplink.PendingDeepLinkManager
 import xyz.ksharma.krail.core.log.log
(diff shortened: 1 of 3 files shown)


ksharma-xyz:
* **#1594** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1594?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1594?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1624, opened 2026-05-22
fix(ios): stop SQLITE_BUSY crash from multiple SQLite driver pools

ksharma-xyz:
The iOS app crashed with `SQLiteExceptionErrorCode: database is locked`
(SQLITE_BUSY) when selecting a theme during the first-launch GTFS import.

Root cause: the same `krailSandook.db` was opened by several independent
`NativeSqliteDriver` connection pools — each repo (`RealSandook`,
`RealNswStopsSandook`, `RealNswBusRoutesSandook`) called
`factory.createDriver()`, and both platform DI modules declared an extra
`single<RealSandook>` with its own factory. Separate pools cannot serialize
through SQLDelight's single-writer borrow, so they race for the file write
lock. sqliter already defaults to WAL + a 5s busy timeout, but the ~5.5s
first-launch stops insert outlives that timeout, so a competing writer (theme
selection) throws. Android masks the same bug via its busy handler + WAL.

Fix, in four layers:
- Share one driver: all repos now consume the shared `KrailSandook` from Koin;
  removed the duplicate `single<RealSandook>` in both platform modules. One
  pool = writes serialize on the Kotlin-side borrow, no cross-pool lock race.
- iOS driver: set WAL explicitly and raise busyTimeout to 30s (defence in depth).
- De-dup the import: `Mutex` in NswStopsManager and NswBusRoutesManager so the
  concurrent RealAppStart + IntroViewModel callers don't both run the full insert.
- Resilient theme write: wrap clearTheme/insertOrReplaceTheme in runCatching so a
  transient lock degrades gracefully instead of crashing.

Verified: detekt, :feature:trip-planner:ui:testAndroidHostTest, and iOS
simulator compile of :sandook and :io:gtfs all green.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/feature/trip-planner/ui/src/commonMain/kotlin/xyz/ksharma/krail/trip/planner/ui/themeselection/ThemeSelectionViewModel.kt
+++ b/feature/trip-planner/ui/src/commonMain/kotlin/xyz/ksharma/krail/trip/planner/ui/themeselection/ThemeSelectionViewModel.kt
@@ -14,7 +14,7 @@ import kotlinx.coroutines.launch
 import xyz.ksharma.krail.core.analytics.Analytics
 import xyz.ksharma.krail.core.analytics.AnalyticsScreen
 import xyz.ksharma.krail.core.analytics.event.trackScreenViewEvent
-import xyz.ksharma.krail.core.log.log
+import xyz.ksharma.krail.coroutines.ext.launchWithExceptionHandler
 import xyz.ksharma.krail.sandook.Sandook
 import xyz.ksharma.krail.sandook.SandookPreferences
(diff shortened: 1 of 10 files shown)


ksharma-xyz:
* **#1624** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1624?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1624?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1663, opened 2026-06-06
fix(search-stop): remove IME padding from floating top bar in single-pane

ksharma-xyz:
SearchTopBar in single-pane was using applyImePadding=true (the default).
This adds windowInsetsPadding(ime) as BOTTOM padding to the bar's Row,
inflating topBarHeightDp by the keyboard height. SearchStopListContent's
padding(top = topBarHeightDp) then included the keyboard height even though
the root Box's imePadding() already excluded that area — double-counting.
Visible as a large gap between the search bar and results on iOS portrait
(where ime.bottom ≈ 336dp); landscape and Android are unaffected because
the keyboard either doesn't report a non-zero ime inset or the gap is less
noticeable.

The top bar is at the TOP of the screen; the keyboard is at the BOTTOM —
they never overlap, so IME padding on the bar serves no purpose. Both
single-pane and dual-pane now pass applyImePadding=false. Root container's
imePadding() handles keyboard avoidance for all children.

### DIFF
--- a/feature/trip-planner/ui/src/commonMain/kotlin/xyz/ksharma/krail/trip/planner/ui/searchstop/SearchStopScreen.kt
+++ b/feature/trip-planner/ui/src/commonMain/kotlin/xyz/ksharma/krail/trip/planner/ui/searchstop/SearchStopScreen.kt
@@ -638,6 +638,12 @@ private fun SearchStopScreenSinglePane(
                 }
                 onTextChange(value)
             },
+            // The top bar floats at the TOP of the screen; the keyboard is at the BOTTOM.
+            // They never overlap, so IME padding on the bar is wrong: it inflates topBarHeightDp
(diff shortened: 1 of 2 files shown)


ksharma-xyz:
* **#1664** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1664?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1663** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1663?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1663?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1688, opened 2026-07-03
fix(trip-planner): collapse same-route legs split by a quick walk

ksharma-xyz:
NSW sometimes represents a bus continuing on a different scheduled
trip (e.g. reversing at a nearby stop) as 3 legs - transport, a
trivial standalone walk, transport again on the same route number -
even though Google Maps and Opal both show it as a single bus leg.
KRAIL was rendering all 3+ legs verbatim.

Collapses [TransportLeg A, walk, TransportLeg B] into one leg when
the walk is a genuine standalone crossing (IDEST) under 2 minutes
and both transport legs share a route number - deliberately not
comparing trip/vehicle IDs, since those legitimately differ in the
real-world case this fixes (different tripCode, opposite-direction
description). Only affects the UI leg list; raw map data and
origin/arrival time resolution are untouched.

Regression test decodes a real captured API response (trimmed of
coords/pathDescriptions only) rather than a hand-built object tree,
so it also catches NSW API shape drift, not just logic regressions.
Two further tests pin the heuristic's boundaries (walk too long,
different route number) with synthetic data, clearly labelled as
such since the real capture doesn't contain those variants.

Also fixes the test suite's json parser, which claimed to mirror
production's Ktor config but was missing isLenient = true (which
production actually sets) - real NSW payloads send some String
fields as unquoted numbers, invisible until a fixture happened to
include one.

Full investigation, reasoning, and what was deliberately left out of
scope: docs/investigations/NSW_715_WALK_LEG_INVESTIGATION.md, linked
from collapseSameRouteQuickWalks()'s KDoc and from CLAUDE.md's
"Per-feature UX rule docs" section so the reasoning stays
discoverable for whoever next touches this code.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01GYdoRHMBgy9sBbfh7Xj1UB

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -221,3 +221,7 @@ that contradicts the doc should also update the doc in the same change.
   landscape (per-screen dual-pane behaviour, compact-height adaptations, breakpoint contract).
 - `docs/POLLING_LIFECYCLE.md` — WhileSubscribed polling rules: `repeatOnLifecycle(STARTED)`
   pattern, why plain `LaunchedEffect` breaks background gating, all polling flows listed.
+- `docs/investigations/NSW_715_WALK_LEG_INVESTIGATION.md` — why `TripResponseMapper.kt`'s
+  `collapseSameRouteQuickWalks()` merges same-route-number legs split by a trivial walk;
+  read before changing leg-merge/split logic in `TripResponseMapper.kt` or
+  `TripResponseLegMapper.kt`.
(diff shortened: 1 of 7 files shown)


ksharma-xyz:
* **#1688** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1688?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1688?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ksharma-xyz:
Test verified works well.

| Before Multiple Legs | Now Same Leg |
|--------|--------|
| Screenshot_20260704-164341 | Screenshot_20260704-164331 |

PR #1693, opened 2026-07-07
feat(timetable): header stop tap opens leg-scoped edit search

ksharma-xyz:
## Story A3 — Timetable header tap opens edit-stop search

Analytics showed 424 taps by 116 users (27% of actives) on the timetable header stop names expecting to change the stop; the tap only opened a departure-board sheet, and the most common follow-up was bailing out to search.

### What changed

- Tapping the origin or destination name in the timetable header now opens stop search pre-scoped to that leg ("Change origin" / "Change destination").
- Picking a stop reloads the timetable in place with the changed leg. The other leg is untouched and there is no back-stack rebuild (reuses the reverse-trip reload mechanics via a shared `reloadWithNewTrip`).
- The departure-board sheet is still available: each header row gains an explicit clock icon that opens it. The feature has users, so it moved rather than died.
- Reverse button unchanged.
- The result travels on a dedicated `TimetableStopChangedResult` bus channel so it can never race SavedTrips' `StopSelectedResult` listener.

### Analytics

- `timetable_stop_header_click` keeps firing with `isOrigin`, plus new param `action: "edit_search"` (pre-fix events carry no `action`, so the dashboard can split before/after).
- New event `timetable_departures_icon_click` (`stopId`, `stopName`, `isOrigin`).

### Tests

- Header-click analytics tests updated for the `action` param.
- New VM tests: departures-icon event, origin change keeps destination, destination change keeps origin, same-stop pick is a no-op, `isTripSaved` recomputed for the new pair.
- `testAndroidHostTest` 370 tests green, detekt clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1694, opened 2026-07-07
feat(timetable): one-tap save-trip prompt at the aha moment

ksharma-xyz:
## Story A2 — Save-trip prompt at the aha moment (plain variant)

Only 25.8% of new users save a trip in week 1, and the aha moment (first timetable load) passes with no save nudge. This adds the one-tap prompt.

Plain variant only: the "Save as commute? Home and Work" variant depends on story A1 (label assignment), which is not built yet.

### What changed

- Users with ZERO saved trips see a "Save this trip?" tile the moment a timetable loads — eligibility resolves synchronously inside the load's own state update, so the tile renders together with the loading state and journeys fill in below it (no layout shift).
- Rendered with the shared InfoTile component ("Save" primary CTA, "Not now" dismiss), matching every other actionable tile. Insets aligned with the OriginDestination header; top gap equals the header-to-actions gap (20dp).
- Tapping Save: haptic tick (same LongPress haptic as label pills), trip saved, and the title-bar star plays a one-shot celebration — 360 spin (intro-screen motion language), gold flash, bouncy size pop, settling back filled. Connects the tile action to the permanent save control.
- Frequency rules in `TimeTableViewModel`, dismissal counts in Sandook `KrailPref`:
  - only while the user has no saved trips at all — first save retires the nudge forever
  - at most once per app session
  - never again for a pair after 2 dismissals
- Star save, reverse, or trip switch clears the prompt; eligibility re-evaluates on the next load.

### Analytics

- New events: `save_trip_prompt_shown` (`variant: plain`) and `save_trip_prompt_action` (`accepted: true|false`, `variant`, `dismissCount` on dismissals). Accept and dismiss share one event to conserve the Firebase 500-event budget.
- `save_trip_click` gains `source: star | prompt` (historical events carry no source; treat null as star).

### Tests

VM tests: prompt visible while loading and stays after journeys arrive (+ shown event), no prompt when the pair is saved, no prompt when ANY other trip is saved, accept saves and fires both events with `source=prompt`, dismiss persists count and fires dismissed event, 2 prior dismissals suppress in a new session, once-per-session across pair changes.

`testAndroidHostTest` green, detekt clean. Manually verified on device (prompt timing, star celebration, haptic, spacing).

### DIFF
--- a/core/analytics/src/commonMain/kotlin/xyz/ksharma/krail/core/analytics/event/AnalyticsEvent.kt
+++ b/core/analytics/src/commonMain/kotlin/xyz/ksharma/krail/core/analytics/event/AnalyticsEvent.kt
@@ -355,24 +355,23 @@ sealed class AnalyticsEvent(val name: String, val properties: Map<String, Any>? 
         }
     }
 
-    /** User tapped Save on the "Save this trip?" prompt. */
-    data class SaveTripPromptAcceptedEvent(val variant: String) : AnalyticsEvent(
-        name = "save_trip_prompt_accepted",
-        properties = mapOf(PROP_VARIANT to variant),
-    )
-
     /**
-     * User dismissed the "Save this trip?" prompt.
+     * User acted on the "Save this trip?" prompt. One event for both outcomes
+     * (accept and dismiss) to stay within the Firebase 500-event budget.
      *
+     * @param accepted true when the user tapped Save, false when dismissed.
      * @param dismissCount Total dismissals for this origin-destination pair
      * including this one — the prompt stops appearing for the pair at 2.
+     * Always 0 when [accepted] is true.
      */
-    data class SaveTripPromptDismissedEvent(
+    data class SaveTripPromptActionEvent(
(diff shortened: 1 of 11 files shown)


ksharma-xyz:
* **#1695** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1695?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1694** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1694?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1694?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1693** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1693?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #1699, opened 2026-07-11
feat(searchstop): unset pill enters choose-mode, row tap assigns label (A1)

ksharma-xyz:
## Summary

Replaces the star + modal-sheet flow for assigning custom labels to stops with an inline expand-in-place flow (`StopLabelAssignRow`), and adds a full-screen Manage labels experience (`ManageStopLabelsScreen`) to replace the old `ManageStopLabelsSheet`.

- Unset pill on a stop row assigns the label instantly; a pill already used on another stop opens a Replace/Remove confirm; an assigned stop collapses to a "Change label / Remove" summary instead of showing the full label wall.
- Manage screen: accordion single-row-expand, chevron instead of pencil/edit icons, per-label preset/hashed icon colors, key-based drag reorder, protected `Home` label can't be renamed.
- `stop_label_stop_assigned` analytics event gains a `source` param (`choose_mode` | `star_sheet`) to split the two assignment entry points on the dashboard.

## Bug fixes bundled in

- Config-change crash: `ManageStopLabelsRoute` was never registered in `SerializationConfig`'s polymorphic `NavKey` module.
- Silent no-op when renaming/creating a label with a name already taken — now shows `ConfirmLabelActionSheet` instead of pretending it saved.
- Drag-reorder in Manage screen: `MoveLabelToIndex` now reorders by label key instead of raw `LazyColumn` index — the reorder-mode banner and unset-label rows share the list with the draggable set labels, so the old index-based move landed items in the wrong slot.
- Stop-label pill height drifting from the transport-mode icon at non-default font scales.
- Ripple/tap target on search-result rows not covering the full padded row height.
- `src/androidUnitTest` in this module was never wired into the Gradle build (only `androidHostTest` is) — three Compose UI test files sat there silently unexecuted. Moved into `androidHostTest`, added the missing `ui-test-manifest` dependency, fixed assertions that had bit-rotted since they were written.

## Deleted (superseded by the new flow)

`AddLabelBottomSheet`, `SaveStopAsLabelSheet`, `LabelConflictSheet`, `SearchStopLabelDialogs` — replaced by `AssignNewLabelSheet` + a now-3-way `ConfirmLabelActionSheet`. Also removed two dead-code paths with zero production callers left, both pre-existing debt from earlier UX rounds (v2/v3):
[...]

PR #1700, opened 2026-07-12
fix(searchstop): label name max length, Manage-button gating, chip height

ksharma-xyz:
## Summary

Follow-up polish on the stop-label assign/manage flow shipped in #1699:

- **Label name max length**: capped at 20 chars (`LABEL_NAME_MAX_LENGTH`), enforced both at the `TextField` (`maxLength`) and inside `normaliseLabelName` itself — the latter is the actual DB-write path (`CreateLabel`/`RenameLabel` call it before persisting), so the cap holds even for a caller that skips the UI.
- **`ManageStopLabelRow`'s rename field** was missing the emoji/punctuation keystroke filter that `AssignNewLabelSheet` already had; both now share one `filterLabelNameInput` function instead of two copies of the same predicate.
- **Manage button no longer shows when no label is set yet**. `shouldShowPillRow` only checked `recentStops`/results existence, not whether any label was actually set — since Home/Work are always seeded (unset) on install, `stopLabels` was never empty, so a fresh install showed a floating "Manage" button with no pills next to it.
- **Manage button height mismatch fixed**: `chipButtonSize()`'s text style was silently falling through to `titleSmall` (18sp line-height) instead of `labelLarge` (20sp) that the pills use — the 2sp gap was the visible height mismatch. `chipButtonSize()`'s own doc comment already said it should match pill geometry; `buttonTextStyle()`'s mapping just never had a case for it.
- Fully-qualified `Text(` call in `StopLabelAssignRow.kt` replaced with a proper import.

## Also investigated (no change needed)

**Spaces in label names**: confirmed safe. `StopLabels.label` is `TEXT NOT NULL PRIMARY KEY`; SQLite/SQLDelight handles UTF-8 and spaces natively, all queries use bound params (no injection surface), and matching is exact-string throughout. The only real risk was normalization consistency, which #1699 already fixed.

## Refactor note

Adding the keystroke filter to `ManageStopLabelRow`'s rename field pushed `ExpandedRenameContent` over the `CyclomaticComplexMethod` threshold. Extracted the Save-button branching into a pure `resolveRenameSaveAction` function (own unit test file) rather than suppressing — per project convention, zero tolerance on suppressing that rule.

## Test plan

- [x] `./gradlew :feature:trip-planner:ui:testAndroidHostTest :feature:track:ui:testAndroidHostTest :taj:testAndroidHostTest --continue` green
- [x] `./gradlew detekt --continue` green, no auto-corrects needed
- [x] Manual: type an over-length / emoji-heavy label name in both Assign and Manage-rename flows, confirm capped/filtered live
- [x] Manual: fresh install (or clear all label assignments) — confirm no floating Manage button until a label is actually set
- [x] Manual: eyeball Manage button vs pill height side by side

### DIFF
--- a/feature/trip-planner/ui/src/androidHostTest/kotlin/xyz/ksharma/krail/trip/planner/ui/searchstop/SearchStopScreenInteractionTest.kt
+++ b/feature/trip-planner/ui/src/androidHostTest/kotlin/xyz/ksharma/krail/trip/planner/ui/searchstop/SearchStopScreenInteractionTest.kt
@@ -243,7 +243,7 @@ class SearchStopScreenInteractionTest {
         composeRule.setContent {
             PreviewTheme {
                 SearchStopScreen(
-                    searchStopState = SearchStopFixtures.recentWithDefaults(),
+                    searchStopState = SearchStopFixtures.recentWithHomeSet(),
                     onEvent = {},
                 )
             }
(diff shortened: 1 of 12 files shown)


ksharma-xyz:
* **#1700** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1700?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1700?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Issue #1697, opened 2026-07-10
Explore: address/POI search via NSW stop_finder + location-aware nearby stops on SearchStopScreen


ksharma-xyz:
## Why

`SearchStopScreen` search is 100% local DB today (`RealStopResultsManager` → `Sandook`/SQLDelight, no network call) — fast, reliable, offline, but only matches transit stops, not street addresses or POIs. NSW's `stop_finder` API can return address/POI hits (`type_sf=any`) and is already partially wired in `RealTripPlanningService.stopFinder()`, but it's dead code today — zero production callers. Separately, GPS/location is fully built (`core/maps`: `UserLocationManager`, `NearbyStopsRepository`) and already drives the map view, but none of that reaches the plain search list — the empty-state always shows 4 hardcoded stops regardless of where the user is.

Citymapper and Google Maps both lead with "don't tell us where you are, we already know" — nearby stops/destinations with zero typing, and free-text search that blends stops + addresses + POIs into one experience. KRAIL has most of the plumbing for this already built; it just isn't surfaced on the primary search screen.

## stop_finder API reference (confirmed against code)

Endpoint: `GET /v1/tp/stop_finder`

| Param | Value | Notes |
|---|---|---|
| `outputFormat` | `rapidJSON` | required |
| `coordOutputFormat` | `EPSG:4326` | required |
| `type_sf` | `any` | **the key param** — default in code today is `StopType.STOP` (transit-only); needs `ANY` passed explicitly (enum already has it) |
| `name_sf` | `<query>` | URL-encoded search text |
| `TfNSWSF` | `true` | NSW API-gateway flag |
| `version` | `10.2.1.42` | present in `StopFinderRequestParams` but **currently commented out** in `RealTripPlanningService.stopFinder()` — needs uncommenting to match the working `/trip` call's pattern |

Response `type` field: `singlehouse` (exact address, `isBest:true` when query is "number + street + suburb") / `street` (street-level only) / `poi` / `stop`. Partial-text queries still return candidates but ranking gets noisier — full "number + street + suburb" is the reliable case.

## Current-state findings

- Search: local-only, fast, in `RealStopResultsManager` — no network call anywhere in this path.
- `stopFinder()`: coded, untested in prod, zero callers.
- Location: `UserLocationManager` + `NearbyStopsRepository` (local-DB haversine query, already distance-sorted) fully working, but scoped to the map view only.
[...]

PR #1701, opened 2026-07-12
feat(searchstop): add address/POI search via NSW stop_finder (Phase 1)

ksharma-xyz:
## Summary

Adds address/POI search to the stop-search screen, behind a remote-config flag (default off). Previously the "Addresses & places" NSW `stop_finder` call existed in code but was never wired up — this reconnects it as its own independent search pipeline, sitting below the existing local (GTFS) stop results.

- New `SearchStopState.ListState`/`SearchResult` handling renders address/POI hits in a separate "Addresses & places" section, each row showing a pin icon + address type subtitle (`AddressSearchListItem`).
- Gated by `SEARCH_STOP_ADDRESS_SEARCH_ENABLED` (Firebase RC, default off) — transit-stop search behavior is completely unaffected when the flag is off; the address pipeline has its own debounce/job/loading state so it can't slow down or interfere with local search either way.
- New Debug Settings toggle ("Address Search") lets developers flip the flag locally without waiting on RC propagation, matching the existing Trip Tracking override pattern. The toggle now reads live (not a one-time snapshot), so flipping it takes effect immediately without recreating the ViewModel.
- Fixes a "No match found" regression where the empty-state message rendered above address results whenever local search had zero hits, even if the API found address matches for the same query — it now only shows when *both* sources are empty.
- Includes investigation notes (`docs/investigations/ADDRESS_ORIGIN_TRIP_INVESTIGATION.md`) from chasing down why a selected address produced no Timetable journey — root-caused to a client-side gap that PR #1702 addresses (NSW flags some first/last-mile footpath legs `footPathInfoRedundant`, which was silently dropping the only walk-leg text telling the rider to walk before boarding).

Ref: https://github.com/ksharma-xyz/KRAIL/issues/1697

## Test plan

- [x] `./gradlew detekt` clean (forced non-cached run)
- [x] `./gradlew compileDebugSources` clean (forced non-cached run)
- [x] `./gradlew :feature:trip-planner:ui:testAndroidHostTest` green
- [x] Manually verified address search results render and selecting one navigates correctly

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/core/remote-config/src/commonMain/kotlin/xyz/ksharma/krail/core/remoteconfig/RemoteConfigDefaults.kt
+++ b/core/remote-config/src/commonMain/kotlin/xyz/ksharma/krail/core/remoteconfig/RemoteConfigDefaults.kt
@@ -160,6 +160,10 @@ object RemoteConfigDefaults {
                 first = FlagKeys.ENABLE_PROTO_BFF.key,
                 second = false,
             ),
+            Pair(
+                first = FlagKeys.SEARCH_STOP_ADDRESS_SEARCH_ENABLED.key,
+                second = false,
+            ),
         )
     }
 }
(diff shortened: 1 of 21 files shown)


ksharma-xyz:
* **#1702** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1702?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1701** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1701?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1701?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1715, opened 2026-07-15
feat(analytics): stop sending raw search query text

ksharma-xyz:
search_stop_query and stop_selected both carried the raw typed query. With
address search shipping, that text can be a street address tied to a pseudo
user ID, which contradicts the privacy policy's no-PII commitment. Replace
the text with queryLength plus a random searchSessionId that joins the two
events per query instance.

One narrow carve-out remains for fuzzy-matcher diagnostics: a query that
returned zero results everywhere, contains no digits, and is 25 chars or
less is still sent in the existing "query" property, so the
top-zero-result-queries dashboard keeps working (SearchQueryAnalyticsRedaction).

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1716, opened 2026-07-15
feat(analytics): fire search_stop_query from the address pipeline

ksharma-xyz:
The address pipeline previously emitted no analytics at all: API error rate,
zero-result rate, and call volume were invisible in the field, and failure is
deliberately silent in the UI. Each resolved, non-stale fetch now fires
search_stop_query with resultSource=address and the same searchSessionId as
the local pipeline's firing, so the two can be joined per query instance.
Cache hits do not fire - the count matches real network calls, which is what
the stop_finder cost model cares about.

This is also the carve-out site for address-eligible queries: only here are
both pipelines' result counts known, so a query like "townhall" that
returned nothing anywhere stays diagnosable while a recognised address never
carries its text.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

### DIFF
--- a/core/analytics/src/commonMain/kotlin/xyz/ksharma/krail/core/analytics/event/AnalyticsEvent.kt
+++ b/core/analytics/src/commonMain/kotlin/xyz/ksharma/krail/core/analytics/event/AnalyticsEvent.kt
@@ -166,6 +166,9 @@ sealed class AnalyticsEvent(val name: String, val properties: Map<String, Any>? 
      * @param resultsCount    Result count on success.
      * @param isError         True when the search pipeline threw.
      * @param zeroResultQuery Raw text only under the redaction carve-out, else null.
+     * @param resultSource    Which pipeline resolved: local stop search or the remote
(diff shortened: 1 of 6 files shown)


ksharma-xyz:
* **#1717** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1717?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1716** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1716?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1716?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1715** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1715?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #1766, opened 2026-07-23
feat(saved-trips): hide Saved Trips section when nothing is saved

ksharma-xyz:
## What

With nothing saved, the home screen rendered a "Saved Trips" sticky header over an empty
section and a placeholder row, pushing Park & Ride down the screen. This drops the whole
section when `savedTrips` is empty. Park & Ride is unchanged and still always renders, so
the add card stays reachable before a first trip is saved.

Also retitles the Add Park & Ride picker description, which overstated what the list holds.

## Changes

- `SavedTripsScreenSections.kt`: `savedTripsListBody` now renders `savedTripsContent` only
  when `savedTrips` is not empty. The `"Saved Trips"` sticky header and the
  `"Tap ★ on a trip to save it here."` row are gone.
  - The placeholder row goes with the header rather than staying on its own. Without a
    header above it, it sat directly under the `"Park & Ride"` sticky header and read as
    Park & Ride copy.
  - `parkRideSection` is untouched and still called outside the branch, so the header, any
    cards, the add card and its hint all render in the empty case.
- `AddParkRideListSections.kt`: picker description changes from
  `"Pick any Sydney car park and its live parking shows on your home screen."` to
  `"Pick a Park & Ride and its live spaces show on your home screen."` The list only holds
  Park & Ride facilities that report live space counts, not street or commercial parking.
- `SavedTripsScreen.kt` / `SavedTripsScreenSections.kt`: drops the now-unused
  `pageHorizontalPadding` parameter on `savedTripsListBody`, its call site, the `Dp` import
  and a `val dim` that no longer had a reader.
- New `SavedTripsListBodyTest.kt` (7 tests): section visibility in the empty, populated and
  loading states, plus the add-card click callback.
- `placeholder = "Search"` on the picker's search field is left alone. It is generic and
  does not make a claim about scope.
- The tracking card still lives inside `savedTripsContent`, so a rider tracking a journey
  with zero saved trips does not see it. That behaviour is unchanged by this PR (the old
  empty branch skipped it too) and is left for a separate change.

## Testing

- `./gradlew :feature:trip-planner:ui:testAndroidHostTest` green, including the 7 new tests
- `./gradlew detekt --rerun-tasks` clean, no auto-corrected files
- `./gradlew compileDebugSources` green
- Installed on a Pixel 10 Pro and checked the empty home screen and the Add Park & Ride
  picker

## Snapshots

Screenshots were reviewed on device during implementation but are not embedded here.

https://claude.ai/code/session_01KJVN8QSe5fNYH5hGo6SEfe

### DIFF
--- a/feature/trip-planner/ui/src/androidHostTest/kotlin/xyz/ksharma/krail/trip/planner/ui/savedtrips/SavedTripsListBodyTest.kt
+++ b/feature/trip-planner/ui/src/androidHostTest/kotlin/xyz/ksharma/krail/trip/planner/ui/savedtrips/SavedTripsListBodyTest.kt
@@ -1,0 +1,163 @@
+package xyz.ksharma.krail.trip.planner.ui.savedtrips
+
+import androidx.compose.foundation.lazy.LazyColumn
+import androidx.compose.foundation.lazy.rememberLazyListState
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.snapshots.SnapshotStateMap
+import androidx.compose.runtime.mutableStateMapOf
+import androidx.compose.runtime.remember
+import androidx.compose.ui.test.assertIsDisplayed
+import androidx.compose.ui.test.junit4.createComposeRule
+import androidx.compose.ui.test.onNodeWithText
+import androidx.compose.ui.test.performClick
+import com.github.takahirom.roborazzi.RobolectricDeviceQualifiers
+import kotlinx.collections.immutable.persistentListOf
+import org.junit.Rule
+import org.junit.Test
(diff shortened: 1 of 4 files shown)


ksharma-xyz:
* **#1766** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1766?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1766?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1775, opened 2026-07-24
refactor: make DI-bound implementation classes internal

ksharma-xyz:
## What

Tightens visibility on implementation classes that are bound behind an interface via Koin (`single<Interface>{}`/`factory<Interface>{}`) within their own module, from `public` (the Kotlin default) to `internal`. First half of #1772.

## Changes

- 31 classes across `core/app-review`, `core/analytics`, `core/app-start`, `core/app-version`, `core/share`, `core/app-info`, `platform/ops`, `gtfs-static`, `io/gtfs`, `discover/network/real`, `feature/park-ride/network`, `feature/trip-planner/network`, `feature/trip-planner/ui`, `sandook`, and `composeApp` changed from `class Foo` to `internal class Foo`.
- Each candidate was found by locating its Koin DI binding, then verified by grepping the exact class name across the whole repo to confirm every reference lives inside its own module (Koin wiring plus same-module tests, both of which resolve `internal` symbols fine). No production or test code outside a class's own module referenced it by concrete type.
- Left untouched: classes already `internal` (`RealSandook`, `RealUserLifecycleStore`, `RealAppReviewDebugSignal`, etc.), classes without an interface indirection (ViewModels, Composables, state/event classes), and `BffEndpointResolver` (public and deliberately referenced by concrete type from several modules' constructors, not the impl-behind-interface pattern this sweep targets).

## Testing

- `./gradlew compileDebugSources` (Android) and `compileKotlinIosSimulatorArm64` (iOS) both green
- `./gradlew detekt testAndroidHostTest --continue` green across the whole repo

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1772, opened 2026-07-24
Enforce internal/private visibility on implementation classes (+ custom lint rule)


ksharma-xyz:
## Problem

Implementation classes bound behind interfaces are declared `public` (the Kotlin
default) in several modules, so they leak out of their module even though callers
should only ever depend on the interface. Only some impls follow the intended
pattern (e.g. `RealAppReviewDebugSignal` is `internal`), so it is inconsistent.

Interfaces should be `public`; their implementations should be `internal` (or
`private` where file-scoped). DI wiring lives in the same module and binds via the
interface type, so tightening the impl to `internal` is non-breaking.

## Known offenders (app-review, from a recent audit)

- `RealAppReviewManager` (commonMain) - `class` should be `internal`
- `AndroidAppReviewRequester` (androidMain) - `class` should be `internal`
- `IosAppReviewRequester` (iosMain) - `class` should be `internal`

These are the ones already spotted; the task is a repo-wide sweep, not just these.

## Scope

1. Sweep every module for implementation classes that are effectively public but
   only referenced via an interface + DI within their own module; make them
   `internal` (or `private` where they never leave the file).
2. Leave the interfaces public.
3. Add a custom detekt rule that flags a `public` class implementing a module
   interface when it is only referenced within its own module, so this cannot
   regress.
4. Wire the rule into `./scripts/fullQualityChecks.sh` and CI detekt so QA runs it.

## Notes

- Verify DI (Koin) still resolves after each change: bindings are by interface type,
  so `internal` impls remain reachable from the same module's `module {}` block.
- KMP test source sets see `internal` of the module under test, so existing unit
  tests keep compiling.

## Acceptance

- No public implementation class that is only consumed via its interface within its
  module.
- Custom detekt rule present and failing the build on violation.
- `fullQualityChecks.sh` and CI run the rule.

ksharma-xyz:
Sweep in #1775, custom detekt rule + CI wiring in #1776. Both merged.

PR #1776, opened 2026-07-24
feat(detekt): add PublicImplementationClass custom rule

ksharma-xyz:
## What

Adds a custom detekt rule, `PublicImplementationClass`, that flags a public class implementing an interface when it should be `internal` per this repo's DI pattern (interfaces are the public surface; Koin-bound implementations are module-internal). Closes the remaining scope of #1772 (the manual sweep shipped separately in #1775).

## Changes

- New module `gradle/build-logic/detekt-rules` (plain Kotlin JVM library, sibling to `:convention` inside the `gradle/build-logic` composite build) containing `PublicImplementationClass` (the rule) and `KrailRuleSetProvider` (registers it under a `krail` ruleset via `META-INF/services`).
- `Detekt.kt`: wires `xyz.ksharma.krail.gradle:detekt-rules` into every module's `detektPlugins` configuration. Since `Detekt.kt` runs in the *root* build's project context but `:detekt-rules` lives in the *build-logic* composite build, it's requested by GAV coordinate (not `project(":detekt-rules")`, which only resolves within build-logic's own build) and resolved via Gradle's composite-build dependency substitution — `settings.gradle.kts` now also `includeBuild`s `gradle/build-logic` at the top level (previously only inside `pluginManagement`, which registers plugins but not regular library substitution).
- `config/detekt.yml`: activates `krail: PublicImplementationClass`, with an `excludeSuperTypes` denylist for external contracts (`Parcelable`, `KoinComponent`, Compose's `Shape`/`Node`/`DrawModifierNode`, etc.) and a path exclude for `:core:testing`'s `Fake*` classes, which are the repo's intentionally-public shared test doubles.
- Fixed 4 more genuine instances of the pattern that the rule caught beyond the prior manual sweep: `Navigator` (`composeApp`), `AndroidAppInfo`/`IOSAppInfo` (`core/app-info`), `DefaultFuzzyStopRanker` (`feature/trip-planner/ui`) — all now `internal`. Narrowing `Navigator` cascaded into `rememberNavigator()`, `rememberPreviewNavigator()`, and `collectEntryProviders()`, which the compiler required to also become `internal` since they exposed `Navigator` in a public signature.
- No changes needed to `scripts/fullQualityChecks.sh` or CI: both already run `./gradlew detekt`, and the rule is wired into that same task for every module.

## Why the rule matches by raw supertype name, not by resolving the interface

Ideally the rule would confirm the implemented supertype is an `interface` declared in this codebase (as opposed to a binary/library contract) using detekt's type resolution (`bindingContext`). In practice, detekt's Kotlin Multiplatform support does not populate a usable binding context for this project's module structure — verified directly: an `@RequiresTypeResolution` implementation of the same rule produced zero findings against a deliberately broken sample class, on both the experimental `detektAndroidMain` task and the `detektMetadataCommonMain` task, while the identical rule logic correctly flagged the same code in a unit test harness with type resolution set up explicitly.

The rule instead matches by raw supertype name, with one purely syntactic distinction that needs no resolution: a supertype written as `Foo(args)` is always a superclass constructor call (extending an abstract/open class, e.g. `ViewModel()`), never interface implementation, so it's excluded outright. What's left after that (`: Foo` or `: Foo by x`) is genuinely ambiguous without resolution, which is what the `excludeSuperTypes` denylist is for.

## Testing

- `gradlew :detekt-rules:test` (12 unit tests) green
- `./gradlew detekt --continue --no-build-cache --rerun` across the whole repo: zero `PublicImplementationClass` findings (confirmed with a fresh Gradle daemon each time — a long-lived daemon caches `detektPlugins` jars by URL and silently serves stale rule bytecode across runs otherwise, which cost significant time to track down during development)
- `./gradlew compileDebugSources` (Android) and `compileKotlinIosSimulatorArm64` (iOS) both green
- `./gradlew detekt testAndroidHostTest --continue --no-build-cache --rerun` across the whole repo green

### DIFF
--- a/composeApp/src/commonMain/kotlin/xyz/ksharma/krail/navigation/Navigator.kt
+++ b/composeApp/src/commonMain/kotlin/xyz/ksharma/krail/navigation/Navigator.kt
@@ -75,7 +75,7 @@ import xyz.ksharma.krail.taj.theme.KrailThemeStyle
  * @return Navigator instance with theme loaded from database
  */
 @Composable
-fun rememberNavigator(state: NavigationState): Navigator {
+internal fun rememberNavigator(state: NavigationState): Navigator {
     val sandook: Sandook = koinInject()
     val navigator = remember(state) { Navigator(state) }
 
@@ -100,7 +100,7 @@ fun rememberNavigator(state: NavigationState): Navigator {
  * Implements NavigatorBase so feature modules can depend on the interface
  * without circular dependencies.
  */
-class Navigator(val state: NavigationState) : NavigatorBase {
+internal class Navigator(val state: NavigationState) : NavigatorBase {
 
     // ═══════════════════════════════════════════════════════════════════════════════
     // THEME MANAGEMENT (Temporary Coupling - See Migration Plan Below)
@@ -388,5 +388,5 @@ private fun Navigator.stackSummary(): String =
 
 // To be used for Previews
 @Composable
-fun rememberPreviewNavigator(navigationState: NavigationState): Navigator =
+internal fun rememberPreviewNavigator(navigationState: NavigationState): Navigator =
     remember { Navigator(navigationState) }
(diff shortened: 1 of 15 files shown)


ksharma-xyz:
* **#1776** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1776?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1776?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1783, opened 2026-08-04
fix(analytics): sanitize location ids and names in event params

ksharma-xyz:
## What

Firebase silently drops any String event parameter longer than 100 characters: the event
still lands, the parameter is simply missing. EFA address and POI ids
(`streetID:...`, `poiID:...`, `coord:...`) run well past that limit, so `stop_selected`
rows for address selections arrived with no `stopId` and address selections could not be
counted.

Those ids are also not safe to send verbatim. An EFA address id embeds the street, suburb
and postcode as plain text, and the matching `stopName` is the address itself, so sending
either would undo the query redaction the app already does. The same ids reach
`fromStopId` / `toStopId` whenever a trip origin or destination is an address, so the fix
is applied centrally rather than per event.

## Changes

- New `AnalyticsParamSanitizer` in `:core:analytics`:
  - `stopId`, `fromStopId`, `toStopId` keep transit stop ids verbatim; a namespaced or
    over-long id is replaced with a stable FNV-1a 64-bit hash prefixed `addr_`
  - `stopName` becomes `address` when the same event carries an address id
  - any other String value over 100 characters is truncated rather than dropped
  - an id counts as identifying when it contains a colon or exceeds the limit, which is
    structural rather than a prefix list, so an unknown future namespace is redacted by
    default
- `AnalyticsEvent` base class routes every properties map through the sanitizer. The
  constructor parameter is renamed to `rawProperties`; `properties` now exposes the
  sanitized map, so no call site hashes or truncates.
- `:core:analytics` gains `withHostTest {}` and a `commonTest` source set; the module had
  no tests before.
- `docs/ANALYTICS_EVENTS.md` documents the sanitizing rules and the fact that address rows
  predating this change carry no `stopId` at all.
- The hash is deterministic across platforms and launches, so grouping by an address id
  still works while the address text never leaves the device.

## Testing

- `./gradlew :core:analytics:testAndroidHostTest` green (12 new tests: transit ids pass
  through, address/POI/coord ids hash, hash is stable and collision-free across two
  addresses, no address text survives, `stopName` redaction, truncation, non-String values
  untouched, sanitizing applied through the event base class)
- `./gradlew :feature:trip-planner:ui:testAndroidHostTest
  :feature:track:ui:testAndroidHostTest :feature:track:state:testAndroidHostTest
  :composeApp:testAndroidHostTest --continue` green
- `./scripts/fullQualityChecks.sh` green (Android compile, iOS simulator compile, detekt)

No UI change, so no device QA and no snapshots.

### DIFF
--- a/core/analytics/build.gradle.kts
+++ b/core/analytics/build.gradle.kts
@@ -16,6 +16,7 @@ kotlin {
         namespace = "xyz.ksharma.krail.core.analytics"
         compileSdk = AndroidVersion.COMPILE_SDK
         minSdk = AndroidVersion.MIN_SDK
+        withHostTest {}
     }
 
     iosArm64()
@@ -28,6 +29,12 @@ kotlin {
     }
 
     sourceSets {
+        commonTest {
+            dependencies {
+                implementation(libs.test.kotlin)
+            }
+        }
+
         androidMain {
             dependencies {
                 // Firebase BOM needed for GitLive Firebase Android platform dependencies
--- a/core/analytics/src/commonMain/kotlin/xyz/ksharma/krail/core/analytics/event/AnalyticsEvent.kt
+++ b/core/analytics/src/commonMain/kotlin/xyz/ksharma/krail/core/analytics/event/AnalyticsEvent.kt
@@ -21,11 +21,19 @@ private const val PROP_TOTAL_COUNT = "totalCount"
 private const val PROP_LEG_COUNT = "legCount"
 private const val PROP_TRANSPORT_MODES = "transportModes"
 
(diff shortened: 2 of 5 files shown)


ksharma-xyz:
* **#1783** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1783?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1783?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1788, opened 2026-08-05
docs(analytics): record the param caps and how to read window reporting

ksharma-xyz:
## What

Docs only. Records two things that were worked out by hand this week so they are not
re-derived: the analytics param caps, and how to read the new window/fold/pane reporting.

## Changes

- `docs/ANALYTICS_EVENTS.md`, new caps section: params-per-event has room (11 on the widest
  event against a limit of 25, even with `pane` on every event), but the **event-scoped
  custom dimension cap is the binding one** — 118 distinct param names against 50
  registerable dimensions. BigQuery is unaffected since every param lands on the row; in the
  GA4 UI an unregistered param is invisible, which looks identical to absent
- New section on window, fold and pane reporting: an unfold emits two rows not one (count by
  `toFoldState = FLAT`), a wide window that stays `SINGLE` is a finding, `formFactor = PHONE`
  on a Fold model is the used-closed segment, `pane` is last-touch and must not be read as
  "where the tap happened" for passive events, and the width class has five values because
  that is what the layout branches on
- Refreshes the event count in the budget section to the measured 52
- Rewords "params are effectively free" in the checklist, which contradicted the new caps
  section

Mirrors the addendum on the KRAIL-Analytics side so the two do not drift.

## Testing

Docs only, no code touched.

### DIFF
--- a/docs/ANALYTICS_EVENTS.md
+++ b/docs/ANALYTICS_EVENTS.md
@@ -7,9 +7,34 @@ Read this before adding or modifying any event in
 
 Firebase Analytics hard-caps the app at **500 unique event names, forever**. GA never
 lets a name be reclaimed from history — a shipped event name is a permanently spent
-slot, even if the code stops sending it. Budget as of 2026-07-22: 60 events defined in
-code plus ~10 historical names, roughly 430 slots left. Update this count when adding
+slot, even if the code stops sending it. Budget as of 2026-08-05: 52 events defined in
+code plus ~10 historical names, roughly 438 slots left. Update this count when adding
 or removing events.
+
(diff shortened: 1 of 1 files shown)


ksharma-xyz:
* **#1788** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1788?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1788?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1791, opened 2026-08-06
fix(park-ride): stop car parks reading above 100% full

ksharma-xyz:
## What

The home Park & Ride card showed "0 spots available, 135% full" for Gosford, and 101% for
Ashfield. Both numbers come from the same two inputs, but only one was bounded:
`spotsAvailable` was clamped at zero while `percentageFull` was not, so the pair
contradicted itself on the same card.

The cause is upstream data, not arithmetic. The live feed reports occupancy above capacity
often enough to matter, through overflow parking, a stale `spots` total, or bays counted in
occupancy but not in capacity. Confirmed against the live feed while testing:

```
[Park&Ride - Gosford - 8 - 225040] Total spots: 1059, Occupied spots: 1466,
Spots available: -407, Percentage full: 100%
```

1466 cars against 1059 spots is 138% raw. A car park cannot be more than full.

## Changes

- `percentageFull` is clamped to 0..100 on the **write** path, the same way `spotsAvailable`
  already was
- Clamped on the **read** path too: rows already stored above 100 would otherwise keep
  rendering that way until the next successful poll replaced them
- The `-1` loading sentinel is preserved. It means "no reading yet", not an empty car park,
  and clamping it into 0..100 would render a misleading 0%
- `toParkRideUiState()` was building its facilities inline instead of calling
  `toParkRideState()`, and that inline copy is the path the home cards actually render from,
  so a fix applied only to the shared function would have missed the screen showing the
  wrong number. It now calls the shared mapper

Left alone deliberately: a facility reporting **zero** capacity still shows "0% full", the
same contradiction in the opposite direction. Nothing in the wild shows it, and rendering it
as unknown is a separate decision.

## Testing

- `./gradlew :feature:trip-planner:ui:testAndroidHostTest` green. `ParkRideMapperTest` now
  has 13 tests, 10 of them new: 135% and 101% capped, spots and percentage agreeing once
  capacity is exceeded, ordinary and genuinely-full readings untouched, empty car park still
  0, zero capacity not dividing by zero, loading sentinel surviving
- Device: installed on an emulator, added Gosford through the picker, and the card reads
  "0 spots available, 100% full" against the same feed data that produced 135%

## Snapshots

| Screen | Before | After |
|---|---|---|
| Home, Park & Ride card | `0 spots available` / `135% full` | `0 spots available` / `100% full` |

### DIFF
--- a/feature/trip-planner/ui/src/commonMain/kotlin/xyz/ksharma/krail/trip/planner/ui/savedtrips/ParkRideMapper.kt
+++ b/feature/trip-planner/ui/src/commonMain/kotlin/xyz/ksharma/krail/trip/planner/ui/savedtrips/ParkRideMapper.kt
@@ -12,6 +12,20 @@ import xyz.ksharma.krail.trip.planner.ui.state.savedtrip.ParkRideUiState.ParkRid
 import kotlin.time.Clock
 import kotlin.time.ExperimentalTime
 
+private const val FULL_PERCENT = 100
+
+/**
+ * Caps a real occupancy reading at [max] while leaving the negative loading sentinel alone.
+ *
+ * [FULL_PERCENT] is the cap everywhere: a full car park is 100%, the feed can report more
+ * than that, and the UI must never repeat it.
+ *
+ * A card that has never had data carries -1 for every number, which the UI reads as
+ * "still loading". Clamping into a 0..100 range would turn that sentinel into a real
+ * looking 0% and show an empty car park where there is simply no reading yet.
+ */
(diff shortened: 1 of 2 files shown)


ksharma-xyz:
* **#1791** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1791?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1791?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1793, opened 2026-08-06
feat(search): gate address search on local stop match count

ksharma-xyz:
## What

Address/POI eligibility decided on query length alone. This adds a second input the app already has at the decision point: how many on-device stops already match the query.

## Changes

- `AddressSearchEligibility.evaluate` takes `localStopResultCount` and `maxLocalStopsForAddressSearch`, and returns a new `STOPS_ALREADY_SUFFICIENT` gate value when more stops than the max already match.
- Queries of `ADDRESS_SEARCH_LONG_QUERY_LENGTH` (12) characters or more bypass the check. A string that long is usually an address and should still get one even if the stop list is busy.
- `localStopResultCount` is **nullable, and null means "not settled yet"**, not zero. On the keystroke the local pipeline is still inside its own debounce, so the only count available belongs to the previous query, which for incremental typing is a broader result set and would suppress calls that should fire. The check runs at the post-debounce site, on the settled count.
- The post-debounce site now clears the address section when it declines to call. Previously a `return@launch` there left the previous query's addresses on screen; that also applied to a flag flip mid-debounce.
- New remote config key `search_stop_address_max_local_stops`, default `10`, validated `0..50`, falling back rather than clamping, same contract as the length key.
- `STOPS_ALREADY_SUFFICIENT` is deliberately a separate enum value from `BELOW_THRESHOLD`: they are different reasons for not calling, and the follow-up PR reports them separately.
- The local search analytics firing moves into `SearchQueryAnalytics.kt` (`trackLocalSearchResolved` / `trackLocalSearchFailed`). Without it the gate wiring pushes `SearchStopViewModel` past its `LargeClass` limit; no suppressions added.
- `ADDRESS_SEARCH_ELIGIBILITY.md` updated with the gate order, the null semantics, and the remote config table.

Left untouched: the cache lookup still happens before the settled count exists, so a cache hit can serve a query the stop-count gate would later decline. That costs no network call, which is the point of the gate.

## Testing

- `./gradlew :feature:trip-planner:ui:testAndroidHostTest :composeApp:testAndroidHostTest --continue` green. New cases cover the boundary (`count == max` stays eligible), the 12-character escape hatch either side of the boundary, precedence over `BELOW_THRESHOLD` and `DISABLED`, `max = 0`, and the null "not settled yet" path; plus the resolver's in-range, out-of-range, oversized, and malformed cases.
- `./scripts/fullQualityChecks.sh` green (Android compile, iOS simulator compile, detekt).
- Device QA on an Android emulator, debug build, with the remote threshold at 9:

| Query | chars | local matches | gate | address call |
|---|---|---|---|---|
| `fulton p` | 8 | 50 | `BELOW_THRESHOLD` | no |
| `fulton pl` | 9 | 10 | `ELIGIBLE` | yes |
| `central st` | 10 | 13 | `STOPS_ALREADY_SUFFICIENT` | no |
| `central stre` | 12 | 50 | `ELIGIBLE` (escape hatch) | yes |

  Going from an eligible query to a suppressed one clears the address section (checked against a UI dump). Rotated portrait and landscape with results loaded: query text, local results and address section all survive, no `FATAL EXCEPTION`, crash buffer clean. iOS build installs and launches on the simulator.

### DIFF
--- a/core/remote-config/src/commonMain/kotlin/xyz/ksharma/krail/core/remoteconfig/RemoteConfigDefaults.kt
+++ b/core/remote-config/src/commonMain/kotlin/xyz/ksharma/krail/core/remoteconfig/RemoteConfigDefaults.kt
@@ -169,6 +169,10 @@ object RemoteConfigDefaults {
                 second = 6,
             ),
             Pair(
+                first = FlagKeys.SEARCH_STOP_ADDRESS_MAX_LOCAL_STOPS.key,
+                second = 10,
+            ),
+            Pair(
                 first = FlagKeys.IN_APP_REVIEW_ENABLED.key,
                 second = false,
             ),
--- a/core/remote-config/src/commonMain/kotlin/xyz/ksharma/krail/core/remoteconfig/flag/FlagKeys.kt
+++ b/core/remote-config/src/commonMain/kotlin/xyz/ksharma/krail/core/remoteconfig/flag/FlagKeys.kt
@@ -78,6 +78,17 @@ enum class FlagKeys(val key: String) {
     SEARCH_STOP_ADDRESS_MIN_QUERY_LENGTH("search_stop_address_min_query_length"),
 
     /**
+     * Above this many on-device stop matches for the settled query, an address/POI
+     * request is suppressed — a rider already looking at a long, correct stop list is
+     * not stuck, and those calls almost never end in an address being picked. Queries of
+     * 12 characters or more bypass this: a string that long is usually an address.
+     * Bounded integer `0..50`; a missing, malformed, or out-of-range value falls back to
+     * `10` client-side rather than being clamped. See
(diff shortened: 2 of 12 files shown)


ksharma-xyz:
* **#1794** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1794?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1793** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1793?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1793?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1794, opened 2026-08-06
feat(analytics): report the address search gate on search_stop_query

ksharma-xyz:
## What

Adds three params to `search_stop_query` so the address search gate can be tuned from data. A suppressed address call produces no address firing, so there is currently no record of the calls the gate declined, and no way to tell a threshold that went too far from one that is working.

Stacked on #1793, which adds the gate this reports on.

## Changes

- `localResultsCount` on the **address** firing: on-device stop matches for the query, so the gate can be scored directly instead of joining back to the local firing on `searchSessionId`. The value was already passed into `trackAddressSearchResolved`, it just was not put in `rawProperties`.
- `addressSearchGate` on the **local** firing, which happens for every settled query: `DISABLED`, `BLANK`, `BELOW_THRESHOLD`, `STOPS_ALREADY_SUFFICIENT`, `ELIGIBLE`, `CACHE_HIT`. Reported by `.name`, so stable SCREAMING_CASE strings, never an object's `toString()`.
- `queryHasDigit` on the local firing (success and error): whether the query contains a digit. A house number is the cheapest address signal there is. A bool, never the typed text.
- `CACHE_HIT` lives on the analytics enum rather than `AddressSearchGate` because it is not an eligibility outcome: eligibility is a pure decision, a cache hit is why an eligible query still made no network call. It takes precedence when reporting, since the cache is consulted on the keystroke before the settled local count exists, so a query can be both cache-served and later stop-count-suppressed.
- The cache hit is recorded against the search session id rather than a bare flag, so a previous query's hit cannot be read as this query's.
- The gate reported on the local firing is recomputed with the settled local count, which is the same input the address job's post-debounce check sees (350 ms debounce outlasts the local pipeline's 100 ms), so the reported value is the decision that actually gets made.
- Rows added to `docs/ANALYTICS_REGISTRY_HANDOFF.md`; param table added to `ADDRESS_SEARCH_ELIGIBILITY.md`.

No new event name: all three are params on an existing event, per `docs/ANALYTICS_EVENTS.md`. `search_stop_query` now carries 8 params, well inside the per-event cap.

## Testing

- `./gradlew :feature:trip-planner:ui:testAndroidHostTest :composeApp:testAndroidHostTest --continue` green. New cases cover each gate value reaching the event, the digit flag, the address firing's local count, and `CACHE_HIT` on a repeat query.
- `./scripts/fullQualityChecks.sh` green (Android compile, iOS simulator compile, detekt).
- Device QA on an Android emulator, debug build, reading the events off the debug analytics log:

| Query | chars | local | reported gate | address firing |
|---|---|---|---|---|
| `fulton p` | 8 | 50 | `BELOW_THRESHOLD` | none |
| `fulton pl` | 9 | 10 | `ELIGIBLE` | `localResultsCount=10` |
| `central st` | 10 | 13 | `STOPS_ALREADY_SUFFICIENT` | none |
| `12 fulton pl` | 12 | 2 | `ELIGIBLE`, `queryHasDigit=true` | `localResultsCount=2` |
| same query repeated | 12 | 2 | `CACHE_HIT` | none |

  No raw query text on any firing outside the existing zero-result carve-out. No `FATAL EXCEPTION` across two rotations.


ksharma-xyz:
* **#1794** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1794?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1794?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1793** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1793?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #1798, opened 2026-08-07
fix(search): remove the address search stop-count gate and result cache

ksharma-xyz:
## What

Removes the address search stop-count gate and the address result cache. Address/POI eligibility is back to the kill switch, a blank check, and the query-length threshold.

The stop-count gate suppressed the call whenever on-device stop matches exceeded `search_stop_address_max_local_stops` and the query was under 12 characters. That is the wrong signal: `13 hassall` matches dozens of Hassall St bus stops, so the "Addresses & places" section stayed empty for exactly the query that wanted an address, and only appeared once the typed string reached 12 characters.

## Changes

- `AddressSearchEligibility.evaluate` drops `localStopResultCount` and `maxLocalStopsForAddressSearch`; `AddressSearchGate` drops `STOPS_ALREADY_SUFFICIENT` and `ADDRESS_SEARCH_LONG_QUERY_LENGTH` is gone
- Delete `AddressSearchMaxLocalStops.kt` plus the `search_stop_address_max_local_stops` flag key, Remote Config default and DI wiring. A value left set server-side is now ignored
- Delete `AddressSearchCache` and `addressSearchCacheKey`. On screen a cached empty result is indistinguishable from a call that never happened, so every eligible query now reaches the network after the 350 ms debounce. To reduce call volume, raise `search_stop_address_min_query_length` or lengthen the debounce, both of which are visible while typing
- `AnalyticsEvent.SearchStopQuery.AddressGate` drops `STOPS_ALREADY_SUFFICIENT` and `CACHE_HIT` so it mirrors `AddressSearchGate` exactly. Historical rows may still carry the removed values
- Left untouched: the `addressSearchRequestToken` staleness guard (still needed, a late response must not overwrite a newer query's state), `search_stop_address_min_query_length`, and the `localResultsCount` / `queryHasDigit` params
- `ADDRESS_SEARCH_ELIGIBILITY.md` rewritten to record why both were reverted, so neither is reintroduced. `ANALYTICS_REGISTRY_HANDOFF.md` narrowed to match

## Testing

- `./gradlew :feature:trip-planner:ui:testAndroidHostTest` green; `./gradlew testAndroidHostTest --continue` green across all modules
- `./scripts/fullQualityChecks.sh` green: `compileDebugSources`, `compileKotlinIosSimulatorArm64`, `detekt --continue`, no auto-corrected files left uncommitted
- Regression tests added: `AddressSearchEligibilityTest` asserts a short address-shaped query stays `ELIGIBLE`, and `SearchStopViewModelTest` asserts local stop matches still report `ELIGIBLE` and still fire the request
- Device QA on an emulator: `13 hassall` now returns the address list (empty before this change); rotation portrait to landscape dual-pane with results loaded and mid-load, query and results survive, no `FATAL EXCEPTION` in logcat


ksharma-xyz:
* **#1798** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1798?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1798?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1802, opened 2026-08-08
fix(saved-trips): prevent duplicate trip identities

ksharma-xyz:
## What

Fixes a persistent Saved Trips crash caused by two database rows resolving to the same Compose `LazyColumn` key. The change repairs already-affected databases, enforces the identity invariant in SQLite, and documents the cross-layer contract so it is not accidentally broken again.

## Root cause

The affected schema-13 database contained both of these rows for the same Central → Wynyard stop pair:

- legacy ID: `200060-200080`
- current ID: `200060200080`

`SavedTrip.tripId` was not used directly by the UI. Both rows were mapped to a `Trip` whose computed ID concatenated the two stop IDs, so both became `200060200080`. Compose then received the same key twice and threw `IllegalArgumentException: Key "200060200080" was already used` while rendering Saved Trips.

This also explains why deleting and re-saving could make the crash persistent: deletion used the UI's concatenated ID, which did not match the legacy hyphenated database row. The legacy row remained, and saving inserted the concatenated row beside it.

## Fix

- Define saved-trip identity as the ordered `(fromStopId, toStopId)` pair and represent its internal ID as `fromStopId->toStopId`.
- Normalize IDs at the `RealSandook` write boundary so older or mismatched callers cannot reintroduce legacy IDs. The mismatch log is explicitly documented as temporary migration telemetry that can be removed after legacy callers age out; the normalization remains permanent.
- Enforce the identity in the current SQLDelight schema and migrated schema with both:
  - `CHECK (tripId = fromStopId || '->' || toStopId)`, which rejects noncanonical IDs even from external/direct writes;
  - `UNIQUE(fromStopId, toStopId)`, which prevents multiple rows representing the same direction-specific pair.
- Add migration `13.sqm`, which:
  - removes any stale `SavedTripByStops` staging table before rebuilding, so a retry is deterministic after an interrupted or externally replayed migration;
  - rebuilds `SavedTrip` with both identity constraints;
  - keeps the newest row for each duplicate pair (`timestamp DESC`, then `rowid DESC` for a deterministic tie-break);
  - rewrites retained IDs to the canonical pair ID;
  - preserves the retained row's stop IDs, names, timestamp, and sort order;
  - leaves unrelated and reverse-direction trips distinct.
- Keep `FakeSandook` behavior aligned with production storage.
- Continue honoring save-prompt dismissal counters stored under the old concatenated trip ID, avoiding an unrelated prompt-frequency reset for existing users.
- Document the invariant and future-change checklist in `sandook/Migrations.md`, with focused KDoc/comments at the UI identity and Sandook write boundary.

## Existing-user and migration impact

- Existing saved trips remain visible with the same names and ordering; only the internal identifier changes.
- If a database has multiple rows for the exact same origin/destination pair, they are intentionally collapsed to the newest row because they represent the invalid state that triggers the crash.
- Fresh installs create the constrained schema directly.
- Existing installs run the shared SQLDelight migration before Saved Trips are read.
- `sandook/Migrations.md` explicitly says pure SQL `.sqm` migrations are invoked by both platform drivers. Android delegates through `SandookCallback.super.onUpgrade`; iOS uses `NativeSqliteDriver(KrailSandook.Schema, ...)`. No new iOS `AfterVersion` callback is needed because this repair is entirely SQL.

## Prevention contract

The documented invariant requires future identity-format changes to update `Trip.tripId`, `RealSandook`, `FakeSandook`, the current `.sq` schema, persisted preference compatibility, and a numbered migration together. It also requires an executable `KrailSandook.Schema.migrate(...)` test with the historical row shapes and preservation assertions. Stop IDs must never be concatenated without a separator or parsed back out of `tripId`.

## Validation

- `./scripts/fullQualityChecks.sh` — passed after the final documentation/constraint amendment, including Android compilation, iOS simulator compilation, and Detekt.
- `./gradlew testAndroidHostTest --continue` — passed (620 tasks) for the crash fix.
- Focused final regression runs passed:
  - `SavedTripIdentityMigrationTest`, including stale staging-table recovery and direct noncanonical-write rejection against both fresh and migrated schemas;
  - `TimeTableViewModelTest`, including legacy prompt-key compatibility.
- Migration regression reproduces the exact legacy/current duplicate pattern, calls generated `KrailSandook.Schema.migrate(13, 14)`, and verifies deduplication, canonical IDs, retained newest data, unaffected rows, write normalization, and database enforcement.
- Physical Android: Pixel 10 Pro on Android 16, debug app updated in place without clearing data. Its retained database was confirmed at schema 13 before launch; logs showed `Upgrading database from version 13 to 14`; the database header then reported 14; the process remained alive; and four pre-existing Saved Trips rendered without a duplicate-key or SQLite exception.
- iOS: iPhone 17 simulator on iOS 26.4, installed over the affected schema-13 database. Migration reduced the three corrupt rows to two canonical pairs. The full Saved Trips → Timetable → unstar → re-save → back → long-press → delete flow completed without the crash.

## Repository migration-tooling note

`verifyMigrations.set(true)` is configured, but the repository does not check in a SQLDelight baseline `.db` and does not configure `schemaOutputDirectory`, so `:sandook:verifyCommonMainKrailSandookMigration` exits before evaluating any migration and no schema-generation task is registered. This is pre-existing on `main`. The executable generated-schema migration test plus real Android and iOS schema-13 upgrades provide the migration verification for this change.

## Screenshots

Not applicable. There is no visual UI change.

### DIFF
--- a/core/testing/src/commonMain/kotlin/xyz/ksharma/krail/core/testing/fakes/FakeSandook.kt
+++ b/core/testing/src/commonMain/kotlin/xyz/ksharma/krail/core/testing/fakes/FakeSandook.kt
@@ -99,9 +99,16 @@ class FakeSandook : Sandook {
         toStopId: String,
         toStopName: String,
     ) {
-        val existingSortOrder = tripsFlow.value.find { it.tripId == tripId }?.sort_order ?: 0L
+        val canonicalTripId = "$fromStopId->$toStopId"
+        val existingSortOrder = tripsFlow.value
+            .find {
+                it.tripId == canonicalTripId ||
+                    (it.fromStopId == fromStopId && it.toStopId == toStopId)
+            }
+            ?.sort_order
+            ?: 0L
         val trip = SavedTrip(
-            tripId,
+            canonicalTripId,
             fromStopId,
             fromStopName,
             toStopId,
@@ -110,7 +117,10 @@ class FakeSandook : Sandook {
             sort_order = existingSortOrder,
         )
         val current = tripsFlow.value.toMutableList()
-        current.removeAll { it.tripId == tripId }
+        current.removeAll {
+            it.tripId == canonicalTripId ||
+                (it.fromStopId == fromStopId && it.toStopId == toStopId)
+        }
         current.add(trip)
         tripsFlow.value = current
     }
--- a/feature/trip-planner/state/src/commonMain/kotlin/xyz/ksharma/krail/trip/planner/ui/state/timetable/Trip.kt
+++ b/feature/trip-planner/state/src/commonMain/kotlin/xyz/ksharma/krail/trip/planner/ui/state/timetable/Trip.kt
@@ -12,8 +12,14 @@ data class Trip(
     val toStopId: String,
     val toStopName: String,
 ) {
+    /**
+     * Stable storage and Compose identity for this ordered stop pair.
+     *
+     * Keep this format aligned with the Sandook `SavedTrip` constraints and migrations.
+     * A separator is required; concatenating opaque stop IDs can create duplicate list keys.
+     */
     val tripId: String
-        get() = "$fromStopId$toStopId"
+        get() = "$fromStopId->$toStopId"
 
     fun toJsonString() = Json.encodeToString(serializer(), this)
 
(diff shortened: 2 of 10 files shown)


ksharma-xyz:
* **#1802** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1802?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1802?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1844, opened 2026-08-14
feat(trip-planner): AI search as its own screen, dialog on tablets

ksharma-xyz:
## What

Replaces the AI search bottom sheet with a surface of its own: a full screen on phones, a centred dialog on tablets and landscape. Along the way, fixes a resolver bug where a single mentioned place was written into the wrong field, and re-paces the listening waveform.

## Changes

**Input surface** (`feature/trip-planner/ui/.../components/ai/`)
- `AiInputSurface` picks full screen or dialog from `AdaptiveLayoutInfo` and font scale; both wrap the same `AiInputContent`.
- Full screen draws into the host window rather than a `Dialog`, so it inherits the app's edge to edge setup and the gradient reaches both edges. `BackHandler` closes it.
- Wears `CloudGradientBackground` and `TitleBar`, matching SearchStop.
- One state slot at the field's trailing edge: mic at rest, stop while listening, wheel while working. `AiListeningBox` swaps into the field's own box at `TextFieldDefaults.MultiLineMinHeight`, so no state change resizes the surface.
- Adapts to font scale: mic leaves the field at 1.3x, actions become a full width column at 1.8x and the content scrolls.
- Deleted: `AiSearchSheet`, `AiSearchSheetStages`, `AiSheetPieces`, `AiLabelChipRow`, `AiInlineSearchBox`. Label chips go with them; labels still resolve through `StopLabelTextResolver`, only the tap shortcut is gone.

**taj**
- `TextField` takes a `trailingIcon`, reserving its width rather than letting it float over text. The control is inset by 8 on both axes, the text keeps 16, so the gap around the control is even.
- Multiline branch sizes to content (`fillMaxWidth`) with a published `TextFieldDefaults.MultiLineMinHeight`; it previously filled its parent, which took the whole sheet.
- `AiThemeGradientTokens`: theme colour plus a partner 97 to 134 degrees away, third stop computed as the HSV midpoint of the shorter arc so a wide pair does not blend through grey.
- `AiWheelMark` uses a horizontal brush; corner to corner spent both end colours in the empty corners of a circle in a square.
- `AiListeningIndicator` and `AiThinkingIndicator` take colours; the listening indicator can drop its mark.
- `StopIcon` added as path data, like `MicIcon`.
- Waveform re-paced: two sines multiplied instead of `abs(sin)`, which doubled the rate and put a hard corner at every zero crossing.

**Extraction**
- A lone place is treated as the destination unless the wording is about leaving. "let's go home" was filling From with the rider's home station. Both platform prompts say so, and `withSinglePlaceInTheRightField` applies the same rule deterministically after extraction.
- `MicPermissionOutcome` replaces the boolean from the microphone request, separating "refused, can ask again" from "refused, must go to Settings".
- Android recogniser gets explicit silence windows, and the ViewModel caps a session at 20s in case a recogniser ignores them.

Resolving still fills From and To and stops there. Loading a timetable stays the rider's action.

## Testing

- `./gradlew testAndroidHostTest --continue` green, including new `SinglePlaceIntentTest` and a test for the listening ceiling.
- Device: installed on a phone and an emulator. Checked the input state, the keyboard raising and lowering, edge to edge, and back and Cancel dismissal.

Not yet verified: rotation with the surface open, the tablet dialog path, iOS, and real speech end to end on a device with the on-device model downloaded.

## Snapshots

| Screen | Before | After |
|---|---|---|
| AI input | bottom sheet | full screen, new screen |


ksharma-xyz:
* **#1844** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1844?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1844?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1846, opened 2026-08-15
feat(analytics): record which row of the search results the rider picked

ksharma-xyz:
## What

Adds `resultIndex` to `stop_selected`: the zero-based row the picked result was sitting at inside its own section. The event already recorded how many results were on screen, but not which one was taken, so a query answered at the top and a query the rider had to scroll through were indistinguishable.

## Changes

- `AnalyticsEvent.StopSelectedEvent` gains `resultIndex: Int?`, omitted when null.
- `SearchStopViewModel` derives the index from the state it already holds rather than threading an index through the six call sites that fire this event, so the value cannot disagree with what was rendered.
- A stop tapped inside an expanded trip card reports the trip's own row, which is the row the rider had to reach before that stop was on screen.
- Sent raw rather than bucketed: first-versus-second is the question being asked, and a numeric parameter adds no string cardinality.
- Registry row added to `docs/ANALYTICS_REGISTRY_HANDOFF.md` (Pending), and the parameter is listed in `docs/SEARCH_QUERY_TELEMETRY_SPEC.md`.

Deliberately unchanged: the existing `displayedLocalCount` / `displayedAddressCount` buckets stay bucketed. They answer a different question and raw counts would add cardinality without improving it.

## Testing

- `:feature:trip-planner:ui:testAndroidHostTest` green, including three new cases: the second result records row 1, the top result records row 0 and sends the parameter, and a selection with no live query sends nothing.
- Zero is pinned explicitly. A top-result pick is the common case, so a parameter that only appeared when ranking went badly would misrepresent the ranker.
- `detekt --continue` green.

No UI change, so no snapshots.


ksharma-xyz:
* **#1846** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1846?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1846?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1847, opened 2026-08-15
fix(trip-planner): with the feature off, there is no way in

ksharma-xyz:
## What

The AI search feature flag reached the ViewModel but not the UI. With the feature off, the wheel still rendered in the search row, still opened the sheet, and Continue did nothing. This makes the entry point disappear instead.

## Changes

- `AiSearchInputUiState` gains `isFeatureEnabled`, read from the existing flag lambda in the ViewModel's `init`.
- `SearchStopRow` takes `isAiSearchAvailable` and does not compose `AiSheetEntryButton` without it. Threaded through `SavedTripsBottomSearchRow` from `SavedTripsScreen`.
- `OpenInput` now refuses when the flag is off, so a caller that has not been told cannot open a sheet whose only action would be inert.
- `StartOver` and `closeInput` rebuild the state from scratch, so both carry the flag across. Without this, starting over would remove the button the rider had just used.
- `AI_SEARCH_UX.md` failure-mode row 1 updated; it was documented as a known dead button.

Read in `init` rather than via `stateIn(WhileSubscribed)`. This state is read as `uiState.value` in places with no active collector, and a shared flow would hand those callers its initial value instead. `OpenInput` re-reads the flag at the moment it matters, so a value that changes mid-session still removes the entry point.

## Testing

- `:feature:trip-planner:ui:testAndroidHostTest` green, three new cases: no way in with the flag off (state false and `OpenInput` does not open), the way in is present and opens with the flag on, and starting over keeps it.
- `:composeApp:testAndroidHostTest` green.
- `detekt --continue` green.

## Snapshots

No snapshot change: with the flag on, which is what the screenshot tests render, the row is identical. The change is the absence of a button in the flag-off case, which has no baseline.

### DIFF
--- a/feature/trip-planner/ui/AI_SEARCH_UX.md
+++ b/feature/trip-planner/ui/AI_SEARCH_UX.md
@@ -25,7 +25,7 @@ Every way this can fail, what the rider gets, and whether a test holds it in pla
 
 | # | Failure | Rider sees | Test |
 |---|---|---|---|
-| 1 | Feature flag off | **Nothing at all.** Continue does nothing | `submit while flag is off does nothing` — the no-op is tested, the dead button is not fixed |
+| 1 | Feature flag off | **No way in.** The wheel is not drawn | `with the flag off there is no way in`, `with the flag on the way in is there and opens`, `starting over keeps the way in`, plus `submit while flag is off does nothing` |
 | 2 | Model not downloaded, or downloading | "still downloading… try again in a moment" | `model downloadable lands in DOWNLOADING`, `model downloading lands in DOWNLOADING too` |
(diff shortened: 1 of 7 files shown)


ksharma-xyz:
* **#1849** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1849?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1848** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1848?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1847** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1847?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1847?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1848, opened 2026-08-15
feat(trip-planner): carry a chosen departure time through to the timetable

ksharma-xyz:
Stacked on #1847.

## What

A departure or arrival time understood by the AI search sheet ("by 6pm") was parsed into a complete `DateTimeSelectionItem` and then discarded. It now shows on the search row and opens the timetable with it.

The parse was never missing. `AiTripIntentTimeResolver` has always returned a full selection. There was nowhere to put it: the date/time picker exists only on the timetable screen, so a time chosen while the rider is still picking stops had no home, by any route, AI or not.

## Changes

- `SavedTripsState` gains `dateTimeSelectionItem`, set by a new `SavedTripUiEvent.DateTimeSelectionChanged`. The AI resolve effect sends it, including when null, so a fresh sentence with no time clears one left over from the previous.
- `SearchStopRow` draws a chip when a time is set, using the same `SubtleButton` and the same `toDateTimeText()` as the timetable's own time button. Tapping clears it.
- `TimeTableRoute` and `navigateToTimeTable` gain `dateTimeSelectionJson`; `triggerTripSearch` passes it.
- `TimeTableViewModel.initializeTrip` takes the JSON and hands the decoded value to `onLoadTimeTable`, which assigns it in the different-trip branch instead of clearing to null. Every existing caller passes null and keeps today's behaviour exactly.

On the route rather than handed over after navigating, for two reasons. The back stack is serialised, so an app killed in the background comes back to the chosen time rather than to now. And `TimeTableViewModel.dateTimeSelectionItem` is a plain `var` outside `uiState`, which survives rotation but not process death; the route makes it durable without moving that field.

`onLoadTimeTable` takes the selection rather than having it applied after. Applying after would mean two fetches, one for now and one for the time asked for, with the wrong one able to land second.

`TimeTableEntry` passes the route value on every initialise and lets the ViewModel decide. Gating on `hasInitializedTrip` would look right and be wrong: it is `rememberSaveable`, so after process death it restores as true while describing a ViewModel rebuilt from nothing.

## Testing

- `:feature:trip-planner:ui:testAndroidHostTest` green, three new cases: a time on the route survives the new-trip clear, no time still starts at now, and unreadable JSON falls back to now rather than failing (a route serialised by an older build must not take the timetable down).
- `:composeApp:testAndroidHostTest` green, including `NavKeySerializationConfigTest`. The new route field is a nullable `String`, so no `SerializersModule` change was needed.
- `AI_SEARCH_UX.md` failure mode 12 updated, and the remaining work on this path recorded there (chip should open the picker; day grammar; why vague day-parts stay unresolved).

## Snapshots

Not captured. The chip only renders when a time is set, which no screenshot-tested state produces yet, and the row is unchanged without one.

Device QA not yet run: no device was connected. Rotation and process-death behaviour on this path are argued from the code above, not observed.

### DIFF
--- a/feature/trip-planner/state/src/commonMain/kotlin/xyz/ksharma/krail/trip/planner/ui/state/savedtrip/SavedTripUiEvent.kt
+++ b/feature/trip-planner/state/src/commonMain/kotlin/xyz/ksharma/krail/trip/planner/ui/state/savedtrip/SavedTripUiEvent.kt
@@ -36,6 +36,12 @@ sealed interface SavedTripUiEvent {
     data class FromStopChanged(val fromJson: String) : SavedTripUiEvent
     data class ToStopChanged(val toJson: String) : SavedTripUiEvent
 
+    /**
+     * A departure or arrival time understood before a timetable was opened. `null` clears it
+     * back to now, which is what starting a fresh search means.
+     */
+    data class DateTimeSelectionChanged(val dateTimeJson: String?) : SavedTripUiEvent
+
     data object StopTracking : SavedTripUiEvent
 
     data class MoveSavedTripToIndex(val tripId: String, val targetIndex: Int) : SavedTripUiEvent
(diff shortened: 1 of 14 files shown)


ksharma-xyz:
* **#1849** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1849?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1848** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1848?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1848?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1847** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1847?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #1849, opened 2026-08-15
feat(trip-planner): understand named days like tomorrow and friday

ksharma-xyz:
## What

The AI time grammar read clock times and offsets from now, so "friday 6pm" landed on 6pm today or tomorrow, and "tomorrow at 9am" on whichever 9am came next. It now reads named days.

## Changes

- `AiTripIntentTimeResolver` resolves "today", "tonight", "tomorrow", and weekday names including the short forms riders write ("fri", "tues", "weds"). Word-boundary matching, so "sun" is not read out of "sunset".
- A named day overrides the next-occurrence rollover. Without one, an already-past clock time rolls to tomorrow, which is right for a bare "at 8am". With one it must not: the rider said which day.
- Named days resolve dates only. "tonight" gives today's date and no hour; a day with no clock time still resolves to nothing, because a date without a time is not a departure and choosing an hour would be the invented precision this resolver refuses everywhere else.
- Both platform prompts (`AndroidAiTextService`, `AiTextBridge.swift`) now say to keep the day word inside `timeText`. They asked for the time phrase alone, so the day was dropped before the resolver saw it.

## Correction to a note this repo carried

The comment on `StopTextResolver` said the groundwork for offering candidates on a weak match was exposing `FuzzyStopRanker`'s score, which it discards at that boundary. That is wrong on three counts, now written down in `AI_SEARCH_UX.md` so the same plan is not re-derived:

1. The score is not what this path picks from. `RealStopResultsManager` returns exact SQL matches first and only falls back to the ranker when there are few of them, so most results carry no score at all.
2. The score does not carry the safety property. The word-boundary guard in `StopSearchTextResolver` is not a duplicate of the ranker: the ranker scores "work" against *70 Powderworks Rd* at roughly 1.0 through its longest-common-substring signal, which it uses deliberately so a rider reading a list gets loose suggestions. Swapping the guard for a score threshold reintroduces the exact bug the guard exists for.
3. The score does not carry priority. "central" scores *Central Station* and *Central Ave* identically; `prioritiseByRelevance` is what puts the station first.

No code was changed for this. The point of the commit is that the note no longer sends the next reader down that path.

## Testing

- `:feature:trip-planner:ui:testAndroidHostTest` green, seven new cases: tomorrow with a time, a weekday landing on its next occurrence, today's own weekday meaning today, a named day beating the rollover, "tonight" naming the day and nothing else, a day word inside a longer word not counting, and a day with no time resolving to nothing.

Prompt changes are not covered by tests on either platform; they are instructions to an on-device model. The resolver is tested against the strings a model is asked to produce.

## Second commit: found on a device

"10am Monday work" put the rider on tomorrow. The model returned "10am" as the time phrase with Monday swallowed into the place, so no day reached the resolver and an already-passed 10am rolled forward the way a bare time should.

The prompt asks for the day word to stay with the time. A request is not a mechanism. The day is now read from the rider's whole sentence as well, which takes the model out of the decision: a day word is a fact about the text. The extracted phrase still wins when both name a day, and a sentence with no day rolls forward exactly as before. Three more tests, including that exact input.

Also caps a listening session at ten seconds rather than twenty, extending to fifteen when words are still arriving in the two seconds before the ceiling. Twenty left a live microphone running well past the point a rider had finished. The extension is measured on words arriving during that window rather than any words at all, since every session has words in it somewhere. The stop button remains the real control.

Device tested on a Pixel 10 Pro: named days, the arrive-by chip, and the timetable opening at the chosen time all confirmed by hand.

## Snapshots

No UI change in this PR.

### DIFF
--- a/core/ai-text/src/androidMain/kotlin/xyz/ksharma/krail/core/aitext/AndroidAiTextService.kt
+++ b/core/ai-text/src/androidMain/kotlin/xyz/ksharma/krail/core/aitext/AndroidAiTextService.kt
@@ -168,7 +168,9 @@ private fun buildExtractionPrompt(riderText: String): String = """
       false for "leave at" / "leaving around" / "departing at". timeIntent is null if the
       rider gave no time at all.
     - timeIntent.timeText is the time phrase verbatim as the rider said it (e.g. "9am",
-      "6:30pm") — do not convert or resolve it.
+      "6:30pm") — do not convert or resolve it. Include any day word said with it
+      ("tomorrow at 9am", "friday 6pm", "tonight at 7"), because the day is part of when
+      they are travelling and is resolved later from this same string.
     - modeHints lists any transport mode words mentioned (e.g. "train", "bus"), verbatim,
       lowercase. Empty array if none mentioned.
     - If the rider names only ONE place, decide which field it belongs in from the
@@ -182,6 +184,7 @@ private fun buildExtractionPrompt(riderText: String): String = """
     Examples:
     "let's go home" -> {"originText": null, "destinationText": "home", "timeIntent": null, "modeHints": []}
     "leaving home around 9" -> {"originText": "home", "destinationText": null, "timeIntent": {"isArrival": false, "timeText": "around 9"}, "modeHints": []}
+    "town hall to bondi junction by 6pm friday" -> {"originText": "town hall", "destinationText": "bondi junction", "timeIntent": {"isArrival": true, "timeText": "6pm friday"}, "modeHints": []}
 
     Rider's message: "$riderText"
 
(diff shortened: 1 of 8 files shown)


ksharma-xyz:
* **#1849** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1849?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1849?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1848** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1848?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1847** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1847?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #1851, opened 2026-08-15
fix(speech): wait longer before deciding a rider has finished

ksharma-xyz:
## What

The speech recogniser was ending sessions while the rider was still thinking. Both silence windows widen by about a second and a half.

## Changes

- `AndroidSpeechToTextService`: `SILENCE_AFTER_A_COMPLETE_SOUNDING_PHRASE_MILLIS` 1800 to 3000, `SILENCE_THAT_ENDS_THE_SESSION_MILLIS` 2500 to 4000. `EARLIEST_THE_RECOGNISER_MAY_FINISH_MILLIS` unchanged.
- `AiSearchInputViewModel`: `STILL_SPEAKING_WINDOW_MILLIS` 2000 to 4000, to match the recogniser's new tolerance.

The phrase window is the one that caused it. A trip said out loud usually ends on something that sounds finished ("...on Monday morning"), and the recogniser treats a complete-sounding phrase as licence to close much sooner than an unfinished one. A session measured on a device ran 6.4 seconds for roughly 4.5 seconds of speech, which is that 1800ms window behaving exactly as configured.

Erring long is the cheap direction: a rider who has finished has a stop button in front of them and pays nothing for a generous window, while a rider drawing breath loses the sentence and starts again.

## Why the second constant had to move too

`STILL_SPEAKING_WINDOW_MILLIS` decides whether words are still arriving as the app's own ceiling approaches. It must not be shorter than the recogniser's silence tolerance. At 2000 against a 4000 tolerance, a rider who paused at eight seconds would look finished to the ViewModel and unfinished to the recogniser, and the app would be the one cutting them off. The two live in different modules on purpose (this one must not depend on a platform source set), so the relationship is stated in a comment on both sides.

## Testing

- `:feature:trip-planner:ui:testAndroidHostTest` green. The extension test now pauses three seconds rather than one, so it exercises a gap the recogniser itself would still be waiting through.
- Installed on a Pixel 10 Pro. The cut-off was reproduced before the change and the timing above was read from logcat.

iOS is unaffected: these are `RecognizerIntent` extras and have no equivalent in the iOS bridge.

### DIFF
--- a/core/speech-to-text/src/androidMain/kotlin/xyz/ksharma/krail/core/speechtotext/AndroidSpeechToTextService.kt
+++ b/core/speech-to-text/src/androidMain/kotlin/xyz/ksharma/krail/core/speechtotext/AndroidSpeechToTextService.kt
@@ -15,8 +15,18 @@ import kotlinx.coroutines.flow.callbackFlow
 import xyz.ksharma.krail.core.log.log
 
 private const val EARLIEST_THE_RECOGNISER_MAY_FINISH_MILLIS = 3_000L
-private const val SILENCE_THAT_ENDS_THE_SESSION_MILLIS = 2_500L
-private const val SILENCE_AFTER_A_COMPLETE_SOUNDING_PHRASE_MILLIS = 1_800L
+
+// Both widened by roughly a second and a half after a rider reported being cut off while still
+// thinking. The phrase window is the one that does it: a trip said out loud usually ends on
+// something that sounds finished ("...on Monday morning"), and the recogniser treats a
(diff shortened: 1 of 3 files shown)


ksharma-xyz:
* **#1851** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1851?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1851?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1862, opened 2026-08-16
fix(trip-planner): match a rider's own word for a place, and leave send to them

ksharma-xyz:
## What

Three defects found while testing the Ask KRAIL surface on a device.

### "office" did not reach the Work label

`StopLabelTextResolver` compared label strings exactly, so a rider whose label is `Work` said "office by Monday morning" and nothing resolved. Exact matching is there for a good reason — it is what stops "home" capturing "Homebush" — so this widens the **vocabulary**, not the matching.

- `LabelSynonyms` groups `work/office/job/workplace`, `home/house`, `uni/university/campus/college`. Every comparison is still a whole-word exact one
- Two passes, literal before synonym: a rider with both a `Work` and an `Office` label has named two different stops, and "office" has to reach the one they called Office
- The weekend commute rule reads its list from the same groups, so an `Office` label counts as a commute exactly as `Work` does. Two lists would have drifted the moment one gained a word

The group list is deliberately small. A synonym that is nearly right is worse than none: it sends a rider to a stop they did not name and gives them no way to see why.

### Speech submitted without being asked

The `Final` transcript called `submit()` directly. The recogniser deciding it has heard a full sentence is not the rider deciding they have finished saying one, and a mis-heard word was already on its way to a search before they could look at it.

It now fills the field and stops listening. Send stays the rider's to press.

### Words only appeared after speaking finished

The Android service emits partials correctly and sets `EXTRA_PARTIAL_RESULTS`. The ViewModel wrote them to `speechTranscript`, which nothing renders, while the field binds to `typedText`. Partials now land in `typedText`, so the sentence builds as it is spoken.

## Testing

- `LabelSynonymsTest` (8): group interchangeability, no leaking between groups, and that `homebush`, `workshop` and `union square` are still not matches. A rider's invented label compares equal only to itself
- `StopLabelTextResolverTest` (+3): the reported `office` case, literal-wins-over-synonym, and a synonym of a label with no stop pinned to it
- `AiSearchInputViewModelTest`: three rewritten for the new send behaviour, one added for partials appearing mid-sentence. The rewrite re-exposed the `advanceUntilIdle` hazard — it runs virtual time past the listening ceiling, so the session stops itself before the assertion; `runCurrent()` is correct there
- Device QA on a physical Pixel: spoke a trip, watched the words arrive, confirmed nothing submitted until send was pressed

### DIFF
--- a/feature/trip-planner/ui/src/commonMain/kotlin/xyz/ksharma/krail/trip/planner/ui/components/ai/AiGreeting.kt
+++ b/feature/trip-planner/ui/src/commonMain/kotlin/xyz/ksharma/krail/trip/planner/ui/components/ai/AiGreeting.kt
@@ -5,6 +5,7 @@ import androidx.compose.runtime.remember
 import kotlinx.datetime.DayOfWeek
 import kotlinx.datetime.TimeZone
 import kotlinx.datetime.toLocalDateTime
+import xyz.ksharma.krail.trip.planner.ui.search.ai.resolve.LabelSynonyms
 import xyz.ksharma.krail.trip.planner.ui.state.savedtrip.StopLabel
 import xyz.ksharma.krail.trip.planner.ui.state.timetable.Trip
 import kotlin.time.Clock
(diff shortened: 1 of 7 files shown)


ksharma-xyz:
* **#1862** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1862?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1862?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1863, opened 2026-08-16
feat(trip-planner): name the Ask KRAIL screen and drop the duplicate speak button

ksharma-xyz:
## What

Four corrections to the Ask KRAIL surface, all found by using it on a device.

### The app bar carries the title again

It was briefly wordless, on the grounds that the field said "Ask KRAIL" as well and the screen was introducing itself twice. The field now asks **"Where to, and when?"**, so the three lines do three different jobs — what this is, an example, what to type — rather than repeating one instruction. The repetition was the actual complaint, not the title.

### The duplicate "Speak" button is gone

At font scales ≥1.8 a worded `Speak` button rendered below the input bar. It dates from when mic and send were a row *underneath* the field, where a bare glyph among worded controls reads badly at large sizes.

Once both folded into the bar, that button became a second way to do the same thing sitting directly under the first one. The mic inside the bar keeps its full touch target at every font scale, so nothing was lost by removing it.

`ACTIONS_STACK_SCALE` stays — it still decides dialog vs full screen — but it no longer changes the controls.

### Padding between the bar and the keyboard: 16dp → 20dp

16 read as the bar resting on the keyboard. 32 was tried and read as floating away from it.

### The bar is plain `surface` in light mode

The grey dates from a flat background, where white on white had no edge at all. The background is now the drifting cloud field, so a white bar reads as an object sitting on top of the light — which is what it is. Against moving colour the grey only looked dirty.

Dark mode still darkens, because the opposite holds there: `surface` is already near black, and the field's light passes straight through anything that does not sit below it. Both branches resolve to one **opaque** colour rather than a translucent tint, since a tint takes its appearance from whatever happens to be behind it and behind this is a gradient that moves.

## Testing

- `./scripts/fullQualityChecks.sh` green (layout invariants, analytics assumptions, Android compile, iOS compile, detekt)
- `./gradlew :feature:trip-planner:ui:testAndroidHostTest` green
- Device QA on a physical Pixel in both themes, at a large font scale — which is the only place the duplicate button was visible

### DIFF
--- a/feature/trip-planner/ui/src/commonMain/kotlin/xyz/ksharma/krail/trip/planner/ui/components/ai/AiInputBar.kt
+++ b/feature/trip-planner/ui/src/commonMain/kotlin/xyz/ksharma/krail/trip/planner/ui/components/ai/AiInputBar.kt
@@ -44,13 +44,33 @@ import xyz.ksharma.krail.trip.planner.ui.search.ai.AiSearchInputUiState
 
 private val BarCornerRadius = 28.dp
 private const val BAR_DARKEN_DARK = 0.45f
-private const val BAR_DARKEN_LIGHT = 0.09f
 private const val SEND_ENTER_SCALE = 0.5f
 private const val SEND_EXIT_SCALE = 0.35f
 private const val SEND_FADE_MILLIS = 140
 
 // Slower than the entry fade, so the shrink is still visible as the button goes.
 private const val SEND_EXIT_FADE_MILLIS = 200
+
+/**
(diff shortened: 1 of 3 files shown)


ksharma-xyz:
* **#1863** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1863?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1863?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1864, opened 2026-08-16
feat(trip-planner): offer to get the rider home instead of two strangers' stops

ksharma-xyz:
## What

The weekend suggestion was always `"Central to Bondi Junction"` for a rider whose only data is their commute.

The weekend rules disqualified their labels **and** their saved trip, and the ladder fell through to two stations they have nothing to do with — every Saturday and Sunday, deterministically.

## Why not just fall back to the commute

That was the first fix, and a test caught it immediately: it suggested `"Home to Work"` on a Saturday, which is precisely what the weekend exclusion exists to prevent. Undoing one bug with the other is not a fix.

Falling back to **home** uses what the rider actually has, without resurrecting the commute.

## Changes

- `SuggestionShape` adds `HomeBound`: `"Get me home by 9pm"`. Short enough never to trip the length cap, true whether or not they are actually out, and the only line that shows the origin can be **left out entirely** — which the nearby-stop resolver already supports but nothing on screen ever demonstrated
- Used when a situation asks for it, or whenever the day's rules disqualify everything the rider has. Only when a home stop is genuinely pinned; there is no promising a home that is not set
- `allowsCommute` moves onto the situation row, so weekday rows keep the commute and weekend rows do not. It was a boolean threaded through the resolver, which is the wrong place for a rule about what day it is
- **Sunday from 15:00 plans tomorrow**: `"Home to Work by 9am tomorrow"`. The one weekend moment when the commute is what somebody is actually thinking about, and it has to say *tomorrow* or it reads as a trip for tonight
- Evenings past 21:00 turn the clause relative, since "by 9pm" has already gone

### Resulting lines

| When | Line |
|---|---|
| Weekday morning | `Home to Work by 9am` |
| Weekday 15–18 | `Work to Home after 6pm` |
| Weekday 18–21 | `Get me home by 9pm` |
| Sunday 15:00+ | `Home to Work by 9am tomorrow` |
| Weekend, non-commute label | `Home to Gym in 20 minutes` |
| Weekend, only commute data | `Get me home by 9pm` |
| Any day past 21:00 | `Get me home in 20 minutes` |

### Metro pairs with yellow

Purple Drip is a magenta, so the teal theme's AI gradient read as teal and pink and the two ends fought rather than travelled. Measured rather than eyeballed: teal to yellow is **119°**, inside the 97–148 band the token file documents. Teal to green would have been 54°, well inside the muddy zone, so yellow is the pair that works.

## Testing

- 4 new greeting tests, including the exact reported case: only-commute data at a weekend now offers home and never reaches the generic pair
- 3 new situation tests for the Sunday exception, and that no other weekend row allows the commute
- The absolute-clause guard extended to `9pm`, and taught that a clause saying "tomorrow" is exempt from its band — `"by 9am tomorrow"` is legitimately in an evening row
- `./scripts/fullQualityChecks.sh` green, `:taj` and `:feature:trip-planner:ui` host tests green
- Device QA on a physical Pixel

### DIFF
--- a/feature/trip-planner/ui/src/commonMain/kotlin/xyz/ksharma/krail/trip/planner/ui/components/ai/AiGreeting.kt
+++ b/feature/trip-planner/ui/src/commonMain/kotlin/xyz/ksharma/krail/trip/planner/ui/components/ai/AiGreeting.kt
@@ -45,6 +45,11 @@ internal data class AiGreeting(val suggestion: String)
 private val COMMUTE_LABELS = LabelSynonyms.commuteLabels
 
 private const val HOME_LABEL = "home"
+
+// The label is home by definition here, so the sentence says home rather than quoting whatever
+// casing the rider typed. "Get me home" is also what somebody would actually say out loud.
+private const val GET_ME_HOME = "Get me home"
+
 private const val DEFAULT_FROM = "Central"
 private const val DEFAULT_TO = "Parramatta"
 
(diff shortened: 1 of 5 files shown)


ksharma-xyz:
* **#1867** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1867?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1866** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1866?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1865** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1865?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1864** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1864?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1864?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1867, opened 2026-08-16
fix(trip-planner): never match a rider's label word against stop names

ksharma-xyz:
## What

Four defects found testing on a device. Three share one root cause: the app treated a rider's word for a place (`work`, `home`, `office`) as ordinary search text.

### `work` matched "Blacktown Workers Club"

With no Work label set, *"I wanna go to work tomorrow at 9am"* filled the destination with **Blacktown Workers Club**. The label resolver found nothing, the chain moved on, and the stop search matched the letters in "Workers".

A rider who says "work" has never meant a club containing those letters. `LabelWordGuard` wraps the stop search so a label word resolves to **a label or to nothing at all**.

Deliberately a wrapper rather than a rule inside the search: typing "workers club" into an ordinary stop search should still find it. What is wrong is reaching it *through a label word in a sentence*, and that distinction belongs to this chain.

### `work by 9am tomm` found no place at all

To the model, "work" is an ordinary noun, so it read a time and nothing else. `withLabelWordAsDestination` scans the rider's own text for a word matching their own labels and uses it as the destination.

Whole words only, so "homebush" is still not "home". It does nothing for a rider who has not made that label — which is why it correctly stayed silent in the no-Work-label case above. **Nothing here comes from the model**, so `AI_SEARCH_UX.md`'s rule holds: it is the rider's word, from their text, matched to their label.

### The advice was unfollowable

"Try the name as it appears on the sign" is useless for a label word — there is no sign that says Work. The message now names the real route: search for the stop and save it under that name.

### A journey from a bus stop to its neighbour

*"To home by 9pm"* while standing at home filled the origin with the next stop along the road. The same-stop guard did hold — they were two different stops — but the result was still nonsense.

Within walking distance of the destination the origin is now left **blank and editable**. Saying nothing beats saying something untrue.

## Deliberately not done

**No shortcut button on the message.** The obvious destination, "Manage your labels", can only rename, reorder and remove — a label is *created* from the stop search. A button landing a rider on a screen that cannot do what the message just asked of them is worse than no button.

`navigateToSearchStop` takes a `labelKey` and looks like the right target, but whether it accepts a label that does not exist yet is unverified, so the plumbing is out rather than half-right. Recorded in §8 of `ASK_KRAIL_UX.md`.

### Banner styling

Folded in rather than raised separately, because it is the same banner whose copy this PR
rewrites.

- `AlertIcon` (new, in `:taj`): a stroked circle with an exclamation, authored as path data like
  `MicIcon` / `StopIcon` / `SendIcon`. A circle rather than a triangle — a stop that did not
  resolve is something to correct, not a danger
- Top aligned, not centred: these messages run to three lines often enough that a centred icon
  drifts into the middle of a paragraph and reads as punctuation
- Message drops to `bodySmall`. This is the app talking, and it should not compete for weight
  with the sentence the rider wrote
- More room inside the banner, and the gap to the input box goes `spacingM` → `spacingXL`

## Testing

- `LabelWordGuardTest` (5): the Workers Club case, every label word guarded, case and spacing, and that the club is **still findable by its real name**
- `LabelWordFallbackTest` (9): the reported sentence, synonyms, punctuation, that an extraction which already found a place is never second-guessed, and that "homebush" is not "home"
- `LabelledStopLocatorTest` (+2): standing at the destination, and well away from it
- `./scripts/fullQualityChecks.sh` green, feature host tests green
- Device QA on a physical Pixel across all four reported sentences

### DIFF
--- a/feature/trip-planner/ui/src/commonMain/kotlin/xyz/ksharma/krail/trip/planner/ui/components/ai/AiInputContent.kt
+++ b/feature/trip-planner/ui/src/commonMain/kotlin/xyz/ksharma/krail/trip/planner/ui/components/ai/AiInputContent.kt
@@ -136,7 +136,9 @@ internal fun AiInputContent(
         // message names the real route and the shortcut waits until it can point at it.
         state.problemMessage()?.let { AiProblemBanner(message = it) }
 
-        Spacer(modifier = Modifier.height(dim.spacingM))
+        // The banner and the box are two different things and used to sit close enough to read
+        // as one block.
+        Spacer(modifier = Modifier.height(dim.spacingXL))
 
         AiInputBar(
             state = state,
--- a/feature/trip-planner/ui/src/commonMain/kotlin/xyz/ksharma/krail/trip/planner/ui/components/ai/AiInputContent.kt
+++ b/feature/trip-planner/ui/src/commonMain/kotlin/xyz/ksharma/krail/trip/planner/ui/components/ai/AiInputContent.kt
@@ -129,6 +129,11 @@ internal fun AiInputContent(
         // Above the bar, not below it. The bar is pinned to the bottom edge, so a message that
         // grows to three lines would otherwise shove it and its controls upward. Above, it
         // takes room from the empty space, which is what that space is for.
+        // No action button yet, deliberately. The obvious destination, "Manage your labels",
+        // can only rename, reorder and remove — a label is CREATED from the stop search, by
+        // finding a stop and saving it under a name. A button landing a rider on a screen that
+        // cannot do the thing the message just told them to do is worse than no button, so the
+        // message names the real route and the shortcut waits until it can point at it.
         state.problemMessage()?.let { AiProblemBanner(message = it) }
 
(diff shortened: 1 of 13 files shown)


ksharma-xyz:
* **#1867** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1867?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1867?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1866** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1866?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1865** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1865?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1864** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1864?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #1868, opened 2026-08-16
feat(trip-planner): show the resolved trip on the Ask KRAIL screen

ksharma-xyz:
## What

A sentence that resolved closed the Ask KRAIL surface, filled the two fields on the home screen behind it, and left the rider to find the Search button. The stops it found were only visible after the screen that found them had gone, and a half-read sentence looked identical to a correct one: both dropped the rider on the home row to work out for themselves which field had been filled.

The surface now shows what it found, above the bar that produced it, with one action straight to the timetable.

Three commits, because two of them are corrections found on device and the reasoning is worth keeping separate from the build that caused them.

### The result

- `AiResultCard`: the two stops as the home row's own `TextFieldButton`, the shared time chip, and **See timetable**
- `WhenChip` moves out of `SearchStopRow` into its own file now that two surfaces draw it. Its click is optional
- `ResolvedTripIntent.hasWholeTrip` gates the button. There is no timetable to load without both ends
- `SavedTripsEntry` no longer closes the surface, and its exit delay constant goes with it. It still writes both stops and the time into `SavedTripsState`, which is what **See timetable** reads and what makes leaving by back land on a row that agrees with the card

### Correction 1: it closed itself, so the card was never seen

`AiSearchInputViewModel.withResolution` set `isInputOpen = false` on resolve. Stopping the entry from closing the surface was not enough; the ViewModel had already closed it. Every outcome now keeps it open.

### Correction 2: read-only stops, and the bar stays

- The stops were tappable, opening the ordinary stop search. That closed this surface and navigated away, so the card was destroyed and the rider never returned to it: a one-way door dressed up as editing in place
- The result replaced the bar. A rider who was misheard had nothing left to type into, and a half-read sentence became a dead end. The result now sits above the bar and both stay: bar is the input, result is the result
- `TextFieldButton` takes a nullable `onClick`, so read-only is a state of the same component rather than a lookalike. Not the same as `enabled = false`, which dims: a read-only field is at full strength and simply is not a control
- No container. It was a grey panel over the moving cloud field; the pills carry their own surface colour, so grouping comes from spacing, which has three sizes rather than one
- **See timetable** names the screen it opens. "See times" said what happens without saying where the rider lands

## Deliberately not done

**No inline stop correction.** Doing it properly means the stop search inside this surface, which is a much larger change. Until then the bar below and the home row behind are both editable, and neither pretends to be something it is not.

## Known, not fixed here

Two layout defects found while testing this, both older than this branch and both raised separately: the send button is not reachable with the keyboard up and multiple lines of text, and the dialog presentation clips its actions in landscape.

## Testing

- `AiResultCardTest` (9): the gate both ways, the working phase, no invented time chip, tapping a stop does nothing, the bar survives a resolve, and each control reports once. Mutation-checked by relaxing the gate, which fails it
- `AiSearchInputViewModelTest`: the two `isInputOpen` expectations that encoded self-closing now assert the surface stays up. That is the path the card tests cannot reach, and it is where the first defect lived
- `./scripts/fullQualityChecks.sh` green, `:feature:trip-planner:ui` and `:composeApp` host tests green
- Device QA on a physical Pixel in both themes


ksharma-xyz:
* **#1869** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1869?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1868** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1868?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1868?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1869, opened 2026-08-16
fix(trip-planner): keep the input bar's actions inside the bar

ksharma-xyz:
## What

With the keyboard up and more than a couple of lines typed, the Ask KRAIL input bar showed a sentence with no way to send it: the mic and send buttons were gone. Dismissing the keyboard brought them back. Same in landscape, where the surface is short for a different reason.

They were never off-screen. They were inside the bar, laid out below the bar's own bottom edge, and removed by the bar's rounded-corner clip.

### The fix

`AiInputBar`'s `Column` had no weighted child. A `Column` measures its non-weighted children first, against the height it was given, and divides what is left among the weighted ones — so the `TextField` took its natural height up to `maxHeightInLines = 6` and the action `Row` was measured from whatever remained. A child measured past the parent's bound is still laid out, at a position past the parent's edge, and the `clip()` does the rest.

- `TextField` takes `weight(weight = 1f, fill = false)`. The unweighted action row is now measured first and always fits; the sentence scrolls inside what is left. This is the shape every phone messaging field has
- `fill = false` so a short sentence still makes a short bar. With `fill = true` one line would stretch the bar to the full height available

### The test

`AiResultCardTest.shortHost_theBarsActionsStayInsideIt`: the surface rendered at the height a keyboard actually leaves, six lines of text, both controls asserted with `assertIsDisplayed()`.

`assertIsDisplayed()` and not a bounds comparison, deliberately. `getUnclippedBoundsInRoot()` reports where a node was laid out, not whether any of it can be seen, so a clipped node returns perfectly legal on-screen bounds — which is the entire bug. Only a display assertion walks the clip chain.

Mutation-checked both ways: removing the weight fails it, restoring it passes.

### Docs

- `docs/learning/2026-08-16-clipped-inside-its-own-parent.md`. The layout cause was read correctly first time; verification is what cost the session. Three probe failures in a row each looked like a result: a harness seeding the field from the wrong state so nothing overflowed, a bounds assertion that cannot detect clipping, and a `contentDescription` that matched no node in the tree
- `docs/LAYOUT_AND_INSETS.md` section 1.4: `Column` measurement order, the rule that the growable child is the one that takes the weight, and the two-way mutation check for any probe of this kind

## Testing

- `:feature:trip-planner:ui:testAndroidHostTest` green, including the new probe
- `detekt --continue` green, no auto-corrections to commit
- Device QA on a physical Pixel: keyboard up with a long sentence, and at large font scale, both controls present

### DIFF
--- a/docs/LAYOUT_AND_INSETS.md
+++ b/docs/LAYOUT_AND_INSETS.md
@@ -75,7 +75,43 @@ Prefer `weight(1f)` anyway when a child should take the leftover space: it state
 and it keeps working if a sibling is added below. But do not go hunting for this as the cause
 of a misplaced child in a `Column` — it will not be.
 
-### 1.4 One scroller per axis
+### 1.4 Measurement order: what must survive a squeeze has to be measured first
+
+A `Column` measures its **non-weighted children first**, against the height it was given, and
+divides what is left among the weighted ones. A child measured past that bound is still laid
+out, at a position past the parent's bottom edge, and a `clip()` on the parent then removes it
+from the screen. The node is not off-screen and not missing: it is **inside its parent, below
+its parent's edge**.
+
+So in any container holding growable content plus controls, **the growable part is the one that
+gets the weight**, not the controls:
+
+```kotlin
+Column(modifier = Modifier.clip(shape)) {
+    TextField(
+        // Measured after the Row below, from what is left. fill = false so a short
+        // sentence still makes a short bar.
+        modifier = Modifier.weight(weight = 1f, fill = false),
+        lineLimits = TextFieldLineLimits.MultiLine(maxHeightInLines = 6),
+    )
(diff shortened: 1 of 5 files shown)


ksharma-xyz:
* **#1869** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1869?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1869?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1868** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1868?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #1870, opened 2026-08-16
fix(remote-config): give every flag a default, and check that it has one

ksharma-xyz:
## What

Three flags had no row in `RemoteConfigDefaults.getDefaults()`: `AI_SEARCH_INPUT_ENABLED`, `BFF_USE_FOR_TRACK` and `JOURNEY_MAPS_AVAILABLE`.

**A missing row is not the same as a default of `false`, although the two are indistinguishable until a flag wants to default to `true`.** Firebase returns an empty string for a key it has never heard of, `RemoteConfigFlag` classifies that as `StringValue("")` rather than a boolean, and `asBoolean(fallback)` then returns `"".toBoolean()` — false — **without ever reading the fallback it was passed**. So an absent flag is not off by decision, it is off by accident, and the accident is invisible until someone wants one on.

`JOURNEY_MAPS_AVAILABLE` is the case where that mattered: it is meant to default on, and was reading off.

### Defaults declared

| Flag | Key | Default |
|---|---|---|
| `AI_SEARCH_INPUT_ENABLED` | `ai_search_input_enabled` | `false` |
| `BFF_USE_FOR_TRACK` | `bff_use_for_track` | `false` |
| `JOURNEY_MAPS_AVAILABLE` | `maps_journey` | `true` |

`bff_use_for_track` stays off until it is enabled in the console. Note that this row, not `BffGtfsRealtimeRepository`'s `asBoolean(fallback = true)`, is what decides the value: a key with a row here resolves as a real boolean, so the fallback is never reached, and the same goes for the "`true`/unset" wording on `FlagKeys.BFF_USE_FOR_TRACK`. Those two are left as they are and a comment on the row records which one wins, so the disagreement does not have to be re-derived by the next reader.

`maps_journey` is on. Worth flagging separately: **nothing in the app reads `JOURNEY_MAPS_AVAILABLE`.** It gets its declared default here rather than being deleted, since removing a flag is a separate decision.

### The check

`RemoteConfigDefaultsTest` walks `FlagKeys.entries` and fails with the names of any flag missing a default. It found `BFF_USE_FOR_TRACK` and `JOURNEY_MAPS_AVAILABLE` on its first run — neither was known about when this started, which is the argument for a test over a comment. Two further assertions cover the ways this rots later: a key declared twice (whichever row wins is an ordering detail of the array) and a default whose `FlagKeys` entry has since been deleted.

Required `withHostTest {}` on `:core:remote-config`, which had `commonTest` dependencies declared but no test source set.

The empty-string trap itself is written up on `RemoteConfigDefaults`' KDoc, since the test can only say a row is missing, not why an absent row is dangerous.

## Testing

- `:core:remote-config:testAndroidHostTest` green (3 tests). Verified it fails first: it named both unknown flags before they were fixed
- `./scripts/fullQualityChecks.sh` green — Android compile, iOS Simulator compile, detekt

### DIFF
--- a/core/remote-config/build.gradle.kts
+++ b/core/remote-config/build.gradle.kts
@@ -16,6 +16,7 @@ kotlin {
         namespace = "xyz.ksharma.krail.core.remoteconfig"
         compileSdk = AndroidVersion.COMPILE_SDK
         minSdk = AndroidVersion.MIN_SDK
+        withHostTest {}
     }
 
     iosArm64()
--- a/core/remote-config/src/commonMain/kotlin/xyz/ksharma/krail/core/remoteconfig/RemoteConfigDefaults.kt
+++ b/core/remote-config/src/commonMain/kotlin/xyz/ksharma/krail/core/remoteconfig/RemoteConfigDefaults.kt
@@ -6,6 +6,15 @@ import xyz.ksharma.krail.core.remoteconfig.flag.FlagKeys
  * Holds the default values for remote configuration.
  * Add new default values here. These will be used as fallbacks when the remote config is not available
  * due to network or other issues.
+ *
+ * **Every key in [FlagKeys] needs a row here, including the ones whose intended default is
+ * `false`.** A key that is absent from this list is not the same as a key defaulting to `false`,
+ * even though the two look identical today. Firebase returns an empty string for a key it has
+ * never heard of, `RemoteConfigFlag` classifies that as a `StringValue("")` rather than a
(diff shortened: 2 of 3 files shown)


ksharma-xyz:
* **#1870** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1870?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1870?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1877, opened 2026-08-19
ci(snapshot): run Roborazzi verification on macOS and refresh drifted goldens

ksharma-xyz:
## What

Wires Roborazzi snapshot verification into CI, re-records the goldens that had drifted while no job was checking them, and settles the animation clock so the new blocking step is deterministic.

`verifyRoborazziAndroidHostTest` ran in no workflow, despite TESTING.md claiming it ran alongside the host tests. `testAndroidHostTest` captures in Roborazzi's default mode, which records a missing or changed golden and still passes, so a pixel change could ship with no signal. Nineteen `:taj` goldens had drifted by the time this job was written, three of them for components that had been replaced outright.

## Changes

- Adds `.github/workflows/snapshot-verify.yml`, called from `build.yml` in parallel with `code-quality`.
  - Runs on `macos-latest`: the goldens are recorded on a Mac and Robolectric's native rendering is not byte-identical across operating systems.
  - `lfs: true` on the checkout, otherwise the runner compares against LFS pointer files.
  - One step per golden-owning module so they can block independently.
  - Uploads the reference/diff/new composites when either step reports a diff.
- Re-records the drifted `:taj` goldens.
- Adds `PreviewMapStopSelectionPane_Loading` to `excludedPreviewNames` in `TripPlannerUiSnapshotTest.kt` and deletes its three goldens: its indeterminate `CircularProgressIndicator` cannot be captured twice the same way.
- `BaseSnapshotTest.captureScreenshot` (`core/snapshot-testing`) settles the animation clock before every capture: pause the Robolectric choreographer, then idle the main looper for a fixed 2 s, longer than the `KrailTheme` surface transition's 1680 ms worst case. Re-records the 90 `:taj` goldens under that clock.
- TESTING.md documents the verify job.

## Why the settle is in this PR and not a later one

The first CI run of this workflow failed the `:taj` step on one golden, `PreviewAiListeningIndicatorFrozen...light_normal`: a whole-image diff, reference `#E8ECF0` against a captured `#FFFFFF`. That is the glow stage of `KrailTheme`'s staged `colors.surface` transition, the race the workflow comment describes for `:feature:trip-planner:ui` and assumed `:taj` did not have. It has it; it just hits fewer files per run, and which files it hits moves between runs.

A blocking check has to be deterministic at the commit that introduces it, so the capture-side fix moved here from later in the stack. Confirmation that the re-recorded set is the settled point on that clock rather than another point on the same race: the image CI rendered for the file it flagged is byte-identical (sha256 `8c1180160aa29f8ff295cedccf64de7bc26989db6bd7c53485c1281c7850a712`) to the settled golden recorded for it.

## Intentionally left for a follow-up

`:feature:trip-planner:ui` keeps `continue-on-error` here. Its goldens are re-recorded later in this stack, so until then the step reports against pre-settle images. The workflow comment and TESTING.md still describe the harness as unsettled; both are rewritten by the commit that flips that step to blocking, so they are left alone here rather than having two commits fight over the same lines.

## Testing

- `snapshot-verify.yml` exercised through this PR's own CI run; the `:taj` step is green against the re-recorded goldens.
- Pausing the choreographer is what makes a fixed advance safe on previews that animate forever: left running, Robolectric answers `nativeScheduleVsync` without moving the clock, so an endless animation re-arms a frame inside the same idle window and `idleFor` never returns. Paused, each vsync lands on the next frame boundary, capping the window at 125 frames.
- No production UI source touched; the repo's `code-quality` and Android/iOS compile lanes run on this PR.

## Snapshots

No Composable, theme or resource code changes here. The only non-workflow source change is in the snapshot harness. The `:taj` golden PNGs move because they were stale relative to components that had already changed on main, and because they were previously captured mid-transition, so the new images show each preview's settled state rather than a different design.


ksharma-xyz:
* **#1882** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1882?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1881** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1881?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1880** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1880?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1879** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1879?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1878** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1878?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1877** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1877?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1877?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1883, opened 2026-08-19
test(trip-planner): compose the real home entry in a flow-test harness

ksharma-xyz:
## What

Adds `FlowTest`, a Robolectric harness that composes the real navigation entry rather than a hand-built approximation of it, and the first flow test on top of it: a regression test for the Ask KRAIL handoff replaying a stale answer onto the search row.

## Changes

- `flow/FlowTest.kt`: builds the SavedTrips entry through the same `entryProvider` the app uses and renders it with `NavEntry.Content()`, real ViewModels wired by constructor with fakes only at the module edges. It exposes the two lifecycle events that break effects:
  - `leaveAndReturn` disposes the entry and composes it again.
  - `recreate` runs a `rememberSaveable` round trip through `StateRestorationTester`.
- `flow/HomeFlowFakes.kt` and `flow/RecordingTripPlannerNavigator.kt`: the edge fakes and a navigator that records instead of navigating.
- `flow/AskKrailHandoffFlowTest.kt`: resolves a trip through Ask KRAIL, lets the surface hand off and close, takes a manual From pick off the real result bus, then leaves and returns. Against the pre-fix ViewModel it fails with the AI's origin back on the row as a Button node reading "Central Station"; with the phase step-down in place it passes.
- `docs/INTEGRATION_TESTING_PLAN.md`: documents the `FlowTest` API with a usage example, the Robolectric findings behind it, and the rule that a flow test is proved to fail against the defect it pins before it is committed.
- `docs/FEATURE_QUALITY_CHECKLIST.md`: the re-launched-effects section gains a line pointing at the harness, since that is the only layer that can re-run a collector.

## What the harness deliberately leaves out

`NavDisplay` and its decorators, which makes ViewModels Activity-scoped rather than entry-scoped. Both survive the two lifecycle events, which is the property these tests need; the cost is written down in the plan rather than papered over.

`askKrail` sends its three events one composition apart on purpose. `TextField` reports its contents from a `LaunchedEffect` keyed on them, so a batched open plus fill echoes `TypedTextChanged` during the resolve and steps the phase back to `IDLE` by itself, which would hide the defect the test exists to catch.

## Testing

- `./gradlew :feature:trip-planner:ui:testAndroidHostTest` green.
- The Ask KRAIL flow test was run against the pre-fix ViewModel and fails there, so it is proved to pin the defect rather than passing vacuously.

### DIFF
--- a/docs/FEATURE_QUALITY_CHECKLIST.md
+++ b/docs/FEATURE_QUALITY_CHECKLIST.md
@@ -178,6 +178,10 @@ unless the trigger is consumed.
       completes, so a re-launched collector sees nothing to do?
 - [ ] Test: after the action completes, assert the trigger no longer reads as actionable —
       not just that the payload was right.
+- [ ] Flow test: re-launch the effect for real. `FlowTest` (in `:feature:trip-planner:ui`
+      androidHostTest, documented in `docs/INTEGRATION_TESTING_PLAN.md`) composes the actual
+      navigation entry and gives you `leaveAndReturn()` and `recreate()`; a unit test cannot
+      re-run a collector, so this is the only layer that sees the second launch.
 
 Shipped example: the Ask KRAIL dialog wrote resolved stops into the home row while
 `phase == RESOLVED`, and kept the phase after closing. Coming back from the stop-search
--- a/docs/INTEGRATION_TESTING_PLAN.md
+++ b/docs/INTEGRATION_TESTING_PLAN.md
@@ -1,0 +1,234 @@
+# Integration testing plan
+
+Status: first milestone built. `FlowTest` lives in `:feature:trip-planner:ui`'s
+`androidHostTest` source set and runs in the ordinary `testAndroidHostTest` lane. The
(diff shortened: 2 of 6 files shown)


ksharma-xyz:
* **#1885** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1885?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1884** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1884?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1883** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1883?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1883?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1884, opened 2026-08-19
test(trip-planner): pin the search-stop round trip across a recreation

ksharma-xyz:
## What

Adds a second flow test on the `FlowTest` harness: pick a From, recreate the screen, pick a To, and assert both stops are on the row afterwards.

Filling the row takes two separate trips to another screen, so a stop held anywhere but the ViewModel would be lost on the way to the second one, and a recreation is exactly when a rider changes their mind about a stop.

## Changes

- `flow/SearchStopRoundTripFlowTest.kt`: the round trip across a `recreate()`. Both picks arrive on the real `ResultEventBus`, so the entry's own `ResultEffect` has to survive being cancelled and re-subscribed on the same channel.
- `docs/INTEGRATION_TESTING_PLAN.md`: records the new test against its milestone.

## Proving the test is not vacuous

There is no shipped defect for this one to fail against. A `DisposableEffect` probe inside the harness's entry stands in for that proof: it shows `recreate()` disposing and recomposing the entry, rather than the test passing because nothing happened.

## Testing

- `./gradlew :feature:trip-planner:ui:testAndroidHostTest` green.
- `./gradlew detekt --continue` green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1885, opened 2026-08-19
build(detekt): analyse test sources and burn down the findings

ksharma-xyz:
## What

Extends detekt to analyse test sources and burns down everything it found, with no suppressions and no baseline entries.

`configureDetekt` only ever pointed at `commonMain`, `androidMain` and `iosMain`, so every rule in `config/detekt.yml` stopped at the production boundary. A test is code a person has to read and change, and the rules that keep production readable keep tests readable.

## Changes

**Wiring**
- `gradle/build-logic/.../Detekt.kt`: adds the KMP test source sets by their real directory names. `androidHostTest` is what `withHostTest {}` creates, not AGP's `androidTest`.
- `config/detekt.yml`: the seven rules the repo already exempts for test code (`StringLiteralDuplication`, `LargeClass`, `TooManyFunctions`, `SpreadOperator`, `UnsafeCast`, `MagicNumber`, `MaxLineLength`) keep that exemption, with the same globs extended to the KMP directories, since `**/test/**` matches none of them. Those exclude lists become block lists so the next addition is a one-line diff.
- Two further exemptions: `formatting: MaximumLineLength` is ktlint's implementation of `style: MaxLineLength`, which has been exempt for tests all along, but only one of the pair carried the exclude list so the exemption did nothing; both now carry it and a comment saying they move together. `complexity: LongParameterList` is a new exemption, because every hit is a fixture builder taking one named parameter per field with a default on each, and replacing that with a parameter object moves the noise to every call site and hides which fields a test is pinning.

**Findings fixed, 55 across 21 files**
- Complexity, which has no escape hatch in this repo. The three JSON-validation tests that tripped `CyclomaticComplexMethod` each held every rule for one record inline in the loop walking the records; the per-record rules move to private top-level functions in the same file, which is the only extraction that lowers the count, since detekt charges a local `fun`'s branches to its enclosing function. The same move clears `NestedBlockDepth`, `LongMethod` and `ComplexCondition` in those files. `isValidButtonCombo` drops from three returns to one expression.
- Dead code: unused private properties and helpers deleted, keeping the calls whose side effect was the point (`awaitItem` consumes an emission; `decodeBase64Url`'s contract is that it does not throw).
- Test doubles: module-local `Fake*` classes are `internal`, which is what `PublicImplementationClass` asks for and what `commonTest` to `androidHostTest` visibility already allows. Wildcard imports expanded. `IllegalStateException` throws become `error()`. A fake's rigged failure is `IllegalStateException` rather than a bare `RuntimeException`.
- Comments that explained a private property move to the thing they belong to. Five files renamed to match their single declaration.
- Formatting autocorrects applied across test sources, and trailing calls wrapped in chains that were already broken across lines.

**Deletion**
- `RealStopResultsManagerTest.kt`: two tests, both commented out behind a "TODO Write UTs separately", and nothing else. A file like that reads as coverage of the stop-search pipeline and provides none. The module's test count is unchanged at 738 with it gone, which is the proof. `docs/INTEGRATION_TESTING_PLAN.md` records the real gap instead: `RealStopResultsManager` is exercised only indirectly today, end to end by `StopSearchPipelineReproTest` and on its ranker alone by `FuzzyStopRankerTest`, so its flag handling, source choice and recent-search bookkeeping are untested. Git history holds the parked tests for whoever writes the real ones.

## Testing

- `./gradlew detekt --continue` green with test sources in scope.
- `./gradlew testAndroidHostTest --continue` green across every module.

## Note on size

This is larger than the repo's usual per-PR limit because turning a rule set on across test sources produces one mechanical sweep that cannot be split without leaving detekt red in between. The commits inside are separated by kind: wiring, real fixes, exemptions, formatting-only, deletion.


ksharma-xyz:
* **#1885** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1885?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1885?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1884** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1884?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1883** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1883?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #1886, opened 2026-08-19
feat(testing): add testTag selectors for end-to-end drivers

ksharma-xyz:
## What

Adds a stable selector layer for end-to-end drivers. Flows had nothing to select on except visible copy, which makes a flow break on a wording change rather than on a regression.

## Changes

- `exposeTestTagsToUiAutomation()` at the app root, as an expect/actual:
  - Android sets `testTagsAsResourceId`, so each `testTag` surfaces as the accessibility node's `resource-id`.
  - iOS is a no-op, because Compose Multiplatform already publishes `testTag` as `accessibilityIdentifier`.
  - One tag therefore serves both platforms with no per-platform selector table.
- Two constants files, `TripPlannerTestTags` and `DebugSettingsTestTags`, hold the 37 tags so a name cannot drift between the screen that declares it and the flow that uses it. Names describe a node's job, never its label.
- Tags rolled out across saved trips, the search row, search stop, timetable, settings and debug config.
- `SettingsScreen`: gives the six unkeyed `item {}` calls explicit keys, which the repo already requires everywhere else.

## Testing

- `Modifier.testTag` is semantics-only, so none of this participates in layout or drawing. `./gradlew :taj:verifyRoborazziAndroidHostTest :feature:trip-planner:ui:verifyRoborazziAndroidHostTest` is green on both golden-owning modules with the tags in place.
- `./gradlew :feature:trip-planner:ui:testAndroidHostTest :feature:debug-settings:ui:testAndroidHostTest :composeApp:testAndroidHostTest --continue` green.
- `./gradlew detekt --continue` green, no autocorrects left uncommitted.

## Snapshots

Screens are touched, but only with semantics-only modifiers and LazyColumn keys, so there is no rendering change. The snapshot verification job added earlier holds that claim: it compares every golden in both modules against a fresh render and passes on this branch.

### DIFF
--- a/composeApp/src/androidMain/kotlin/xyz/ksharma/krail/TestTagResourceIds.android.kt
+++ b/composeApp/src/androidMain/kotlin/xyz/ksharma/krail/TestTagResourceIds.android.kt
@@ -1,0 +1,8 @@
+package xyz.ksharma.krail
+
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.semantics.semantics
+import androidx.compose.ui.semantics.testTagsAsResourceId
+
+actual fun Modifier.exposeTestTagsToUiAutomation(): Modifier =
+    semantics { testTagsAsResourceId = true }
--- a/composeApp/src/commonMain/kotlin/xyz/ksharma/krail/KrailApp.kt
+++ b/composeApp/src/commonMain/kotlin/xyz/ksharma/krail/KrailApp.kt
@@ -3,6 +3,7 @@ package xyz.ksharma.krail
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.runtime.produceState
+import androidx.compose.ui.Modifier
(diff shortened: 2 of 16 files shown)


ksharma-xyz:
* **#1888** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1888?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1887** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1887?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1886** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1886?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1886?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1887, opened 2026-08-19
test(e2e): add Maestro smoke and nightly flows

ksharma-xyz:
## What

Adds a checked-in Maestro flow suite: six flows split into two lanes, plus a shared step and a README.

`smoke/` is fast enough to gate a pull request; `nightly/` covers process lifecycle and permission denial, where a slow run is fine.

## Changes

- `.maestro/smoke/`: cold launch, the full plan-a-trip journey against the real API, and a rotation sweep across home, search stop and the timetable. Rotation is the one worth having, because it destroys and recreates the Activity, which is this repo's top crash class and is invisible to detekt and unit tests.
- `.maestro/nightly/`: the longer flows, including permissions-denied.
- `.maestro/shared/dismiss-intro.yaml`: the first-run carousel dismissal, kept outside both lanes so a directory run never treats it as a test.
- `.maestro/README.md`: the conventions below, written at each one.
- Selectors are `testTag` ids throughout, so a wording change cannot break a flow and a failure always means behaviour changed. The app id is a parameter passed with `-e`, so one set of flows serves Android debug, Android release and iOS.

## Four conventions, each a fix for a failure this suite hit while being written

- **Wait rather than assert after anything async.** The search row animates in, so asserting it in place passes warm and fails cold: green locally, red on CI's first run only.
- **Scroll to list items instead of asserting in place.** Maestro matches only what is on screen, and the "Save this trip?" prompt pushes journey cards below the fold on a short viewport, so an in-place assert measures device height rather than behaviour.
- **Restore orientation in `onFlowComplete`.** A rotation flow that fails leaves the device in landscape, where the search row collapses to a pill, so one failure cascaded into three.
- **Guard the carousel dismissal on `intro.screen`.** Otherwise it fails on every run after the first.

## Testing

- All six flows verified against an Android emulator and an iOS simulator.
- No production source touched, so unit tests and snapshot verification are unaffected; both lanes still run on this PR.

### DIFF
--- a/.maestro/README.md
+++ b/.maestro/README.md
@@ -1,0 +1,121 @@
+# End-to-end flows
+
+[Maestro](https://maestro.mobile.dev) flows that drive the real app on a real device or
+simulator. They cover the things unit tests and detekt structurally cannot: whether the app
+launches, whether a rider can actually plan a trip, and whether a screen survives having its
+Activity destroyed underneath it.
+
+Verified against **Maestro 2.8.0**. `setOrientation` needs 2.x; on 1.x the rotation sweep will
+not parse.
+
+```sh
+curl -fsSL "https://get.maestro.mobile.dev" | bash
+export PATH="$PATH:$HOME/.maestro/bin"
+maestro --version   # expect 2.x
+```
+
+## Two lanes
+
+| Lane | Runs | Contents |
+|---|---|---|
+| `smoke/` | Every PR | Launch, plan a trip, rotation sweep. Fast enough to gate a merge. |
+| `nightly/` | Nightly cron, `prod/**`, manual dispatch | Process lifecycle and permission-denial cases. Slower, and not worth blocking a PR on. |
+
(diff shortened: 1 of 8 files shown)


ksharma-xyz:
* **#1888** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1888?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1887** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1887?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1887?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1886** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1886?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #1888, opened 2026-08-19
ci(e2e): run Maestro flows on pull requests and nightly

ksharma-xyz:
## What

Runs the Maestro suite in CI: two workflows matching the two flow directories.

## Changes

- `.github/workflows/maestro-pr-smoke.yml`: the three smoke flows on every non-draft pull request. Boots an API 34 emulator, installs the debug APK, runs the lane, and checks logcat for `FATAL EXCEPTION` as a separate step, since a crash in a background coroutine does not necessarily fail a flow. Capped at 20 minutes and deliberately kept out of `build.yml`, so a slow emulator never delays the APK build.
- `.github/workflows/maestro-nightly.yml`: the whole suite at 03:00 AEST, on `prod/**` and on dispatch, across two jobs. Android mirrors the PR lane over all six flows. iOS builds the app with `xcodebuild` against a booted simulator and runs smoke plus nightly there, which works because the same `testTag` ids reach iOS as `accessibilityIdentifier`s with no per-platform selectors.
- TESTING.md: documents both lanes, the `APP_ID` convention, the selector rule and the flake policy.

## Retry policy

Asymmetric on purpose, and documented where it is set. The iOS nightly leg retries once, because simulator runs are flakier and that lane reports rather than gates, so a genuine regression still fails twice. Nothing else retries: a flake that can block a merge should be fixed, not re-run.

## Secrets

This lane cannot use the `-PciQuality=true` placeholder API keys that detekt relies on, because planning a trip asks the real API for real journeys. Real secrets are therefore a hard requirement, so Dependabot is excluded for the same reason as `build-android.yml`.

## Testing

- Verified with `actionlint` under the same `-shellcheck=` configuration `lint-workflows.yml` uses.
- Both flow lanes were run locally against an emulator and a simulator before the workflows were written.

### DIFF
--- a/.github/workflows/maestro-nightly.yml
+++ b/.github/workflows/maestro-nightly.yml
@@ -91,29 +91,43 @@ jobs:
           profile: pixel_6
           disable-animations: true
           emulator-options: -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
+          # Same shape as maestro-pr-smoke.yml, and for the same reason: the
+          # emulator is gone once this script returns, so anything needing adb
+          # has to happen here rather than in a later step.
           script: |
             adb install -r androidApp/build/outputs/apk/debug/androidApp-debug.apk
-            adb logcat -c
-            maestro test -e APP_ID=xyz.ksharma.krail.debug .maestro/
 
-      - name: Check for fatal exceptions
-        if: always()
-        run: |
-          adb logcat -d > logcat-android.txt || true
-          if grep -q "FATAL EXCEPTION" logcat-android.txt; then
-            echo "::error::FATAL EXCEPTION found in logcat"
-            grep -A 30 "FATAL EXCEPTION" logcat-android.txt
-            exit 1
-          fi
-          echo "No fatal exceptions."
+            adb shell settings put global window_animation_scale 0
+            adb shell settings put global transition_animation_scale 0
+            adb shell settings put global animator_duration_scale 0
+
+            adb logcat -c
+
+            set +e
+            maestro test -e APP_ID=xyz.ksharma.krail.debug .maestro/
+            MAESTRO_EXIT=$?
+            set -e
+
+            adb logcat -d > logcat-android.txt || true
+            mkdir -p maestro-output
+            cp -R "$HOME/.maestro/tests" maestro-output/ 2>/dev/null || true
+
+            if grep -q "FATAL EXCEPTION" logcat-android.txt; then
+              echo "::error::FATAL EXCEPTION found in logcat"
+              grep -A 30 "FATAL EXCEPTION" logcat-android.txt
+              exit 1
+            fi
+            echo "No fatal exceptions."
+
+            exit $MAESTRO_EXIT
 
       - name: Upload Maestro debug output
-        if: failure()
+        if: always()
         uses: actions/upload-artifact@v7
         with:
           name: maestro-nightly-android-${{ github.run_id }}
           path: |
-            ~/.maestro/tests/**
+            maestro-output/**
             logcat-android.txt
           if-no-files-found: ignore
           retention-days: 7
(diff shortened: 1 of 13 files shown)


ksharma-xyz:
* **#1888** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1888?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1888?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1887** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1887?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1886** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1886?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

ksharma-xyz:
Holding this PR unmerged. The lane it adds fails on its own first run, and I could not establish a verified fix, so merging it would turn a red 25-minute job on for every pull request.

**What happened** (run 32275014079, job 96140256257)

All three smoke flows failed on the same first assertion:

```
[Failed] 03-rotation-sweep (38s) (Assertion is false: id: savedtrips.screen is visible)
[Failed] 01-launch-home    (36s) (Assertion is false: id: savedtrips.screen is visible)
[Failed] 02-plan-trip      (34s) (Assertion is false: id: savedtrips.screen is visible)
```

Ruled out:

- `APP_ID` is correct. `xyz.ksharma.krail.debug` matches `applicationId` plus the debug `applicationIdSuffix`, and `launchApp` completed on all three flows.
- The tag exists and is wired. `TripPlannerTestTags.SAVED_TRIPS_SCREEN = "savedtrips.screen"`, and `KrailApp` applies `Modifier.exposeTestTagsToUiAutomation()`, whose Android actual sets `testTagsAsResourceId`.
- Animations are already disabled on the runner, so this is not the animator scale.
- The same flow passes locally. Debug APK from this branch installed on an API 36 emulator, `pm clear` first to force the first-run state: `01-launch-home` completes every assertion including the intro-carousel dismissal.

The difference between the two runs is a warm local emulator against a freshly booted CI one. The plausible reading is that `runFlow when visible: intro.screen` is evaluated before the app has drawn its first frame, so the guard skips, the carousel stays up, and the following `extendedWaitUntil` burns its 30 s against a screen that will never appear. Flow durations of 34-38 s fit that: launch, an instant skip, then the full 30 s timeout. That is a hypothesis, not a diagnosis, and the artifact that would settle it was never uploaded, for the reason below.

**Two workflow defects that are confirmed, and are what made this expensive**

1. `Check for fatal exceptions` runs `adb logcat` after the `android-emulator-runner` step has already torn the emulator down. It blocked on `- waiting for device -` from 16:30:06 to 16:42:00 and only stopped when `timeout-minutes: 20` killed the job (`exit code 143`). Any flow failure therefore costs a 12-minute hang. The logcat check needs to run inside the emulator-runner `script:`, where a device still exists.
2. Because the job ended cancelled rather than failed, `Upload Maestro debug output` (`if: failure()`) was skipped. The view hierarchy needed to tell "tag missing" from "wrong screen" is discarded precisely when a flow fails. Defect 1 causes this one.

Fixing 1 and 2 first would make the next run diagnosable rather than another 25-minute guess.

**State of the rest of the stack**

#1886 (testTag selectors) and #1887 (the flow suite) are merged. Nothing in this repo runs those flows yet, so main is not affected by holding this PR; the flows are checked in and runnable by hand.

ksharma-xyz:
Pushed the fix. Three defects, all confirmed rather than guessed:

1. **The flows.** `launchApp` returns once the process is foregrounded, not once Compose has drawn, so on a freshly booted emulator the caller-side `runFlow when visible: intro.screen` was evaluated against an empty hierarchy. It skipped, the carousel stayed up, and the 30 s wait counted down against a screen that was never going to arrive. `shared/dismiss-intro.yaml` now settles first and no-ops by itself; the six callers invoke it unconditionally. `savedtrips.screen` waits raised to 60 s.
2. **The hang.** The logcat check moved inside the emulator-runner `script:`, so it runs while a device still exists. maestro's exit code is preserved across it.
3. **The missing evidence.** Uploads are `always()` now, and the output is copied into the workspace first, because actions/upload-artifact does not expand `~` and would have uploaded nothing even had the step run.

Verified on a local emulator cleared to first-run state, so the carousel is up: 3/3 smoke flows pass in 3m58s, and they pass again warm. actionlint clean.

PR #1891, opened 2026-08-20
feat(detekt): add shrinking-baseline ratchet with lazy-item-key and screenshot-preview rules

ksharma-xyz:
## What

Adds a shrinking-baseline ("ratchet") mechanism to the custom detekt rule set, and the first
two rules built on it: `LazyItemKeyRule` and `ScreenshotPreviewAnnotation`. Both rules fail the
build for any new violation while a plain-text baseline under `config/` grandfathers the
offenders that already exist.

## Changes

- `RatchetBaseline.kt`: shared helper that reads a newline-delimited allow-list from
  `config/<name>-baseline.txt` at the repo root. Same shape as the existing
  `config/test-wiring-baseline.txt`. The file may only shrink: fixing an offender means
  deleting its line in the same change, and when the file empties the rule becomes an
  unconditional gate.
- `LazyItemKeyRule.kt`: flags `item { }` / `items(...)` calls inside a `LazyColumn`,
  `LazyRow`, `LazyVerticalGrid` etc. that pass no `key`, which makes Compose fall back to
  positional identity and breaks scroll-state and animation on reorder or insert. Codifies the
  rule already written in `CLAUDE.md`.
- `ScreenshotPreviewAnnotation.kt`: flags previews that pair `@ScreenshotTest` with a bare
  `@Preview` instead of `@PreviewComponent` / `@PreviewScreen`, which silently captures a
  single configuration and misses dark mode and font scaling.
- Baselines seeded from the current tree: `config/lazy-item-key-baseline.txt` (2 entries),
  `config/screenshot-annotation-baseline.txt` (33 entries).
- `Detekt.kt` passes the repo root through to the rules so a baseline resolves from any module.
- Both rules registered in `KrailRuleSetProvider` and switched on in `config/detekt.yml`.
- Follow-up commit matches lazy-scope receivers by exact name rather than by substring, so a
  type merely containing "LazyColumn" in its name is not treated as a lazy scope.

Deliberately not doing this with detekt's own `baseline.xml`: that file is per-module,
regenerated wholesale by `detektBaseline`, and therefore widened by accident. A hand-edited
text file keeps every grandfathered offender for a rule in one reviewable place.

## Testing

- `gradle/build-logic :detekt-rules:test` green (`LazyItemKeyRuleTest`,
  `ScreenshotPreviewAnnotationTest`, `RatchetBaselineTest`; 3 new test classes)
- No production source changed, so no device QA applies

### DIFF
--- a/config/detekt.yml
+++ b/config/detekt.yml
@@ -104,6 +104,9 @@ exceptions:
     active: true
 
 krail:
+  # Grandfathered offenders: config/lazy-item-key-baseline.txt
+  LazyItemKeyRule:
+    active: true
   PublicImplementationClass:
     active: true
     # Interfaces backed by external/library contracts (not this repo's own module
@@ -112,6 +115,9 @@ krail:
     # :core:testing's Fake* classes are the repo's canonical shared test doubles, deliberately
     # public so other modules' test source sets can construct them directly.
     excludes: ['**/core/testing/**']
+  # Grandfathered offenders: config/screenshot-annotation-baseline.txt
+  ScreenshotPreviewAnnotation:
+    active: true
 
 naming:
   ClassNaming:
--- a/config/lazy-item-key-baseline.txt
+++ b/config/lazy-item-key-baseline.txt
@@ -1,0 +1,14 @@
+# Unkeyed lazy-list item baseline (LazyItemKeyRule)
+#
+# Grandfathered `item { }` calls in a lazy list that pass no key, so Compose identifies the slot
+# by position. The rule fails the build for any NEW one, so this list only ever shrinks.
+#
+# Each line: <path relative to repo root>|<enclosing function name>|<allowed count>
+#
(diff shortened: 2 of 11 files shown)


ksharma-xyz:
* **#1893** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1893?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1892** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1892?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1891** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1891?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1891?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1892, opened 2026-08-20
feat(detekt): ban collectAsState, snackbars, and CyclomaticComplexMethod suppressions

ksharma-xyz:
## What

Adds three more custom detekt rules on top of the ratchet from the parent PR:
`CollectAsStateBan`, `SnackbarBan` and `CyclomaticSuppressBan`. Each encodes a convention that
until now lived only in `CLAUDE.md` prose.

## Changes

- `CollectAsStateBan.kt`: flags `collectAsState()`, which keeps collecting while the screen is
  backgrounded. `collectAsStateWithLifecycle()` is the only accepted form. See
  `docs/POLLING_LIFECYCLE.md`.
- `SnackbarBan.kt`: flags `Snackbar`, `SnackbarHost` and `SnackbarHostState` usage. Snackbars
  are not part of this design system's vocabulary.
- `CyclomaticSuppressBan.kt`: flags any `@Suppress("CyclomaticComplexMethod")` at file,
  class or function level, including the aliased spelling. `CLAUDE.md` sets zero tolerance for
  suppressing that rule, but nothing enforced it.
- `config/cyclomatic-suppress-baseline.txt` seeded with the single pre-existing offender
  (`TripPoller.kt`) so the gate can be switched on without blocking on a refactor. The file is
  closed: entries may be deleted, never added.
- `TripPoller.kt` gains a comment pointing at that baseline entry so the suppression is not
  mistaken for an accepted pattern.
- All three registered in `KrailRuleSetProvider` and switched on in `config/detekt.yml`.

## Testing

- `gradle/build-logic :detekt-rules:test` green (`CollectAsStateBanTest`, `SnackbarBanTest`,
  `CyclomaticSuppressBanTest`; 3 new test classes covering positive and negative cases)
- `./gradlew detekt --continue` green against the whole tree with the three rules active
- No production behaviour changed, so no device QA applies

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1893, opened 2026-08-20
test(taj): measure every colour-token pairing in both schemes

ksharma-xyz:
## What

Adds a contrast test matrix over the fixed `KrailColors` tokens in both colour schemes, plus
structural parity guards on the `KrailColors` data class and a check that brand hex values are
not restated outside `:core:transport`. `ThemeContrastTest` already covered the six
user-selectable theme colours; this covers the other half, where a regression is one hex digit
in `Color.kt` and the surface it lands on is chosen in a different file entirely.

## Changes

- `KrailColorTokenContrastTest.kt`: measures every foreground/background token pairing the app
  renders, in light and dark. Translucent tokens are composited over their background before
  measuring, because `luminance()` ignores alpha; without this, 20%-alpha `outlineSubtle`
  reports about 21:1 against white when the eye sees closer to 1.6:1.
- `KNOWN_FAILURES` records the 6 pairings that do not clear their threshold today, each with
  the ratio measured when it was added and a note on what the pairing actually is (an alert
  container fill, two deviation colours on the dimmed past-departure row, and two hairline
  outlines). Thresholds are never lowered to accommodate one. A pairing that starts passing
  also fails the test, so a fix cannot leave a stale entry behind.
- Self-heal guard: adding a field to `KrailColors` fails the test until someone decides whether
  it carries a contrast obligation.
- `KrailColorsParityTest.kt` (JVM-only, needs `kotlin.reflect`): asserts no `KrailColors`
  constructor parameter is optional, so a new token cannot silently default; and that light and
  dark differ on every field not explicitly listed as deliberately shared.
- `TransportModeContrastTest.kt`: every transport-mode colour clears 3:1 on both surfaces or is
  a recorded exception.
- `BrandHexUniquenessTest.kt`: no production file outside `:core:transport` restates a mode
  brand colour as a literal hex.
- `ThemeContrastTest` now uses the shared `ContrastAnalyzer.TEXT_CONTRAST_AA` constant instead
  of its own private copy of 4.5.
- Three preview-only hex literals replaced with the tokens holding the identical value
  (`train_theme` is `0xFFF6891F`, `metro_theme` is `0xFF009B77`), which is what
  `BrandHexUniquenessTest` now requires. Rendered output is unchanged; only previews and their
  now-unneeded `MagicNumber` baseline entries are touched.
- Removes an unused `LARGE_TEXT_SIZE_CONTRAST_AA` constant in `A11yColors.kt` and its
  `baseline.xml` entry.
- `taj/build.gradle.kts` gains `withHostTest {}` so the JVM-only parity test has a task to run in.

## Testing

- `./gradlew :taj:testAndroidHostTest :feature:trip-planner:ui:testAndroidHostTest :composeApp:testAndroidHostTest --continue` green
- No rendered UI changed (preview-only literal swaps resolve to the same colours), so no
  snapshot diff and no device QA applies

### DIFF
--- a/composeApp/baseline.xml
+++ b/composeApp/baseline.xml
@@ -16,8 +16,6 @@
     <ID>LongParameterList:SplashScreen.kt$( letter: String, animationStarted: Boolean, logoColor: Color, modifier: Modifier = Modifier, fontSize: TextUnit = TextUnit(65F, TextUnitType.Sp), delayMillis: Int = 100, )</ID>
     <ID>LongParameterList:SplashViewModel.kt$SplashViewModel$( private val sandook: Sandook, private val analytics: Analytics, private val appInfoProvider: AppInfoProvider, private val ioDispatcher: CoroutineDispatcher, private val appStart: AppStart, private val preferences: SandookPreferences, private val appVersionManager: AppVersionManager, )</ID>
     <ID>MagicNumber:SplashScreen.kt$0.7f</ID>
-    <ID>MagicNumber:SplashScreen.kt$0xFF009B77</ID>
-    <ID>MagicNumber:SplashScreen.kt$0xFFF6891F</ID>
     <ID>MagicNumber:SplashScreen.kt$1.2f</ID>
     <ID>MagicNumber:SplashScreen.kt$1100</ID>
     <ID>MagicNumber:SplashScreen.kt$1200</ID>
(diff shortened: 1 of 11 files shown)


ksharma-xyz:
* **#1893** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1893?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1893?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1892** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1892?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1891** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1891?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #1894, opened 2026-08-20
feat(detekt): add ClockSystemBan rule and guard the WhileSubscribed register

ksharma-xyz:
## What

Adds `ClockSystemBan`, a detekt rule that keeps new inline `Clock.System` reads out of
production code, and a test that keeps the `WhileSubscribed` register in
`docs/POLLING_LIFECYCLE.md` in step with the source.

### ClockSystemBan

- `ClockSystemBan.kt` flags inline `Clock.System.now()` / `Clock.System` reads in production
  source. A file that reads the wall clock inline cannot be tested at a chosen instant, which is
  why time-dependent logic here is currently tested by waiting or not at all.
- Two shapes are exempt in the rule itself and never need a baseline row: a parameter default
  (`now: Instant = Clock.System.now()`, `clock: Clock = Clock.System`) and the Koin binding in a
  `di` package. Test source sets are exempt entirely.
- `config/clock-system-baseline.txt` grandfathers the reads that exist today as
  `<path>|<count>` rows: 20 files, 57 reads. The count is an allowance, not a target, so a
  listed file cannot quietly grow a new read. Line numbers are deliberately not recorded because
  they churn on unrelated edits.
- The file is shrink-only: counts may be lowered and rows deleted, never raised or added. When
  it empties the rule becomes a zero-tolerance gate.
- Registered in `KrailRuleSetProvider`, switched on in `config/detekt.yml`.

### Polling register guard

- `PollingLifecycleGuardTest.kt` reads `docs/POLLING_LIFECYCLE.md`, scans production sources, and
  fails when the two disagree. `SharingStarted.WhileSubscribed` only stops background work if the
  UI subscribes with lifecycle awareness; when it does not, nothing fails and the app polls the
  network behind the lock screen.
- It sees every `SharingStarted.WhileSubscribed(` in `commonMain` / `androidMain` / `iosMain`,
  attributed to its enclosing member and class, plus any `collectAsState(` that is not
  `collectAsStateWithLifecycle(`. Each flow must appear in the doc's register classified as
  "polls" or "just state"; an unregistered flow and a stale row both fail.
- It guards the register, not the call site: whether a listed flow is actually collected under
  `repeatOnLifecycle` is a Compose call-graph question a text scan cannot answer. The KDoc says
  so explicitly.
- `docs/POLLING_LIFECYCLE.md` restructured into that register so it has a machine-readable shape.
- The scan skips `gradle/build-logic/`. It already skipped its own source, which has to spell the
  banned call out in order to search for it; `CollectAsStateBan` is the same case one step
  removed, since a detekt rule cannot match a string without containing it and neither can its
  test fixtures. Nothing under there is Compose, so the exclusion costs no coverage of app code.

## Testing

- `gradle/build-logic :detekt-rules:test` green (`ClockSystemBanTest`, new)
- `./gradlew :composeApp:testAndroidHostTest` green (`PollingLifecycleGuardTest`, new)
- `./gradlew detekt --continue` green with the rule active against the seeded baseline
- No production behaviour changed, so no device QA applies

### DIFF
--- a/composeApp/src/androidHostTest/kotlin/xyz/ksharma/krail/quality/PollingLifecycleGuardTest.kt
+++ b/composeApp/src/androidHostTest/kotlin/xyz/ksharma/krail/quality/PollingLifecycleGuardTest.kt
@@ -104,11 +104,20 @@ class PollingLifecycleGuardTest {
         return found
     }
 
-    /** This file is skipped: it necessarily spells the banned call out as a search string. */
+    /**
+     * This file is skipped: it necessarily spells the banned call out as a search string.
+     *
+     * So is [BUILD_LOGIC_DIR], for the same reason one step removed. `CollectAsStateBan` is the
+     * detekt rule that bans the call, and a rule cannot match a string without containing it —
+     * nor can its test fixtures. Nothing under there is Compose: it is the build's own static
+     * analysis, so it can never hold a screen that collects state, and excluding it costs this
+     * guard no coverage of app code.
+     */
     private fun kotlinSources(): Sequence<File> =
         repoRoot.walkTopDown()
             .onEnter { it.name != "build" && it.name != ".git" }
             .filter { it.isFile && it.extension == "kt" && it.name != GUARD_SOURCE_FILE }
+            .filterNot { it.invariantPath().contains(BUILD_LOGIC_DIR) }
 
     private fun productionSources(): Sequence<File> =
         kotlinSources().filter { file ->
@@ -125,6 +134,9 @@ class PollingLifecycleGuardTest {
         const val COLLECT_AS_STATE = "collectAsState("
         const val COLLECT_AS_STATE_WITH_LIFECYCLE = "collectAsStateWithLifecycle"
         const val GUARD_SOURCE_FILE = "PollingLifecycleGuardTest.kt"
+
+        /** The included build holding this repo's own detekt rules. See [kotlinSources]. */
+        const val BUILD_LOGIC_DIR = "/gradle/build-logic/"
 
         val PRODUCTION_SOURCE_SETS =
             listOf("/src/commonMain/", "/src/androidMain/", "/src/iosMain/")
--- a/composeApp/src/androidHostTest/kotlin/xyz/ksharma/krail/quality/PollingLifecycleGuardTest.kt
+++ b/composeApp/src/androidHostTest/kotlin/xyz/ksharma/krail/quality/PollingLifecycleGuardTest.kt
@@ -1,0 +1,151 @@
+package xyz.ksharma.krail.quality
+
+import java.io.File
+import kotlin.test.Test
+import kotlin.test.fail
+
+/**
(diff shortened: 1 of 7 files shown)


ksharma-xyz:
* **#1909** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1909?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1908** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1908?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1906** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1906?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1905** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1905?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1904** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1904?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1903** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1903?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1902** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1902?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1901** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1901?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1900** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1900?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1899** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1899?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1898** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1898?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1897** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1897?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1896** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1896?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1895** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1895?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1894** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1894?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1894?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1911, opened 2026-08-21
docs(testing): split TESTING.md into doctrine plus three reference pages

ksharma-xyz:
## What

Splits `TESTING.md` into an entry point plus three reference pages, and fixes the places where
the testing docs had drifted away from what the build actually does.

Docs only: no production or test code changes.

## Changes

- `TESTING.md` keeps the doctrine and becomes the entry point; the detail moves to three pages
  under `docs/testing/`:
  - `LAYERS.md` — what each test layer is for: unit, Compose interaction, flow, snapshot, iOS,
    and end-to-end.
  - `GUARDS.md` — every custom detekt rule and guard test, the ratchet contract they share, the
    Gradle verification tasks, `scripts/check_layout_invariants.py`, and where each `config/`
    file is consumed. There was no single place that listed these, so the usual way to discover a
    guard was to have it fail.
  - `COVERAGE.md` — what the number measures, what is excluded and why, which modules are in the
    aggregate, and what coverage is allowed to gate.
- Drift fixed:
  - `CLAUDE.md` listed three modules as having `withHostTest {}`. It is now most of them, so the
    list is replaced with the `grep` that answers the question, plus a note that `verifyTestWiring`
    fails the build for a module with test sources and no host-test task. A stale list here is
    worse than no list, because it reads as exhaustive.
  - The iOS lane tables described tasks that no longer match the build.
  - The snapshot-mechanism description no longer claimed behaviour the current setup does not have.

## Why this is over 500 lines

It is a restructure: `TESTING.md` loses 373 lines and the three new pages absorb them plus the
material that was previously nowhere. Splitting the split across PRs would mean landing a
`TESTING.md` that points at pages that do not exist yet, or pages nothing links to. The diff is
large but almost entirely moved and reorganised prose.

## Testing

- `./gradlew :composeApp:testAndroidHostTest --continue` green, which is what runs
  `LearningLogActionsTest` over `docs/`
- `python3 scripts/check_layout_invariants.py` and `check_stale_assumptions.py` both clean

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -38,12 +38,24 @@ file" section for the exact row format before adding one.
 
 Modules use KMP `androidLibrary { withHostTest {} }` — this creates `testAndroidHostTest`, not the standard AGP task name.
 
-Modules that have `withHostTest {}` enabled:
-- `feature/trip-planner/ui`
-- `feature/track/ui`
-- `feature/track/state`
+**Most modules have `withHostTest {}` now** — roughly half the build files in the repo, and every
+module that owns a test suite. Do not keep a list here; it goes stale. Check the module's own
+`build.gradle.kts`, or:
 
-If a module is missing `testAndroidHostTest`, add `withHostTest {}` inside its `androidLibrary {}` block in `build.gradle.kts`.
+```sh
+grep -rl "withHostTest" --include=build.gradle.kts .
+```
+
+If a module is missing `testAndroidHostTest`, add `withHostTest {}` inside its `androidLibrary {}`
+block in `build.gradle.kts`. You will not have to notice this yourself: `verifyTestWiring` fails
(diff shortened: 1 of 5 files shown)


ksharma-xyz:
* **#1912** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1912?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1911** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1911?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1911?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1912, opened 2026-08-21
build(coverage): measure the whole repo, exclude what execution cannot prove

ksharma-xyz:
## What

Makes the coverage number mean something: measures the whole repo instead of a handful of
modules, excludes the code where execution is not evidence of testing, deletes verified-dead code
rather than counting it as untested, and adds floors that fail the build on a regression.

Merged line coverage after this change is **63.48%**, measured across 26 registered modules.

### Kover wiring moves into the convention plugin

- `Coverage.kt` applies Kover to every module with production sources and registers it into the
  root aggregate, so `koverHtmlReport` at the root produces one merged report instead of per-module
  ones nobody reads. The list of measured modules is derived from the build rather than
  hand-maintained; 26 modules are now in the aggregate.
- `:core:testing` is skipped: its production code *is* test infrastructure, so a number for it says
  nothing about the app.
- Coverage is host-only. Kover instruments JVM bytecode, so it measures `testAndroidHostTest` and
  has nothing to say about `iosSimulatorArm64Test`; Kotlin/Native is not supported upstream.

### What is excluded, and why

- `@Composable` functions and `ComposableSingletons*`. A Composable is executed by any snapshot
  test that renders it, marking every line covered while asserting nothing beyond "it did not
  crash". Counting them makes the number rise when a preview is added rather than when behaviour
  is checked. What stays measured is code where a wrong branch is a bug you can assert on:
  reducers, mappers, repositories, ranking, date handling.
- Codegen: SQLDelight output, Wire proto classes, `BuildKonfig`, Compose Resources accessors.
- Declarations rather than logic: Koin module blocks, `expect`/`actual` wrappers around Android
  SDK services a host test cannot drive, and the generated members on the `AnalyticsEvent` sealed
  hierarchy.

### SQLDelight codegen package rename

- Generated queries and table classes move to `xyz.ksharma.krail.sandook.db`. They previously
  shared `xyz.ksharma.krail.sandook` with the hand-written repositories, so no package filter could
  separate them and all but one generated file counted as covered production code. The repositories
  above them stay measured. 131 mechanical lines: import updates and `.sq` file moves.

### Dead code deleted

- 901 lines of unreferenced UI removed (`NswTransportLineContrastPreview`, `LineBadgeChip`,
  `TransportModeLineBadge`, `DepartureBoardExt`, `TimePicker`, `FestivalType`, and the snapshot
  goldens for them). Verified unreferenced before deletion. Deleting untested code is the honest
  alternative to reporting it as a coverage gap forever.

### Floors

- Merged line coverage floor 61, plus per-module floors for `:feature:departures:ui` 87,
  `:feature:track:ui` 82 and `:core:date-time` 68 — each set two points under what it currently
  measures (89.12 / 84.54 / 70.27, measured 2026-08-21). They catch a regression such as a deleted
  suite or a large untested surface landing; they are not targets.
- Raising one is a deliberate act in its own commit quoting the new measurement; lowering one
  requires saying what was given up.
- `koverVerify` runs in the code-quality workflow alongside the report.
  `warningInsteadOfFailure` is deliberately not set: a floor that only warns is a comment.
- The HTML report now uploads whenever the step ran, including when `koverVerify` failed, since
  that is exactly the run where someone needs it to see which module lost coverage.

### Codecov

- `codecov.yml` checked in, explicitly informational: nothing there blocks a merge, because a
  second gate computing coverage differently from Kover produces arguments rather than tests. It
  carries a 0.5% drift threshold and adjusts the base for removals, so deleting dead code does not
  register as a win.

### Maestro lanes

- The PR smoke lane keeps `01-launch-home` and `02-plan-trip`; `03-rotation-sweep` moves to the
  nightly lane. The lane retries once (`MAESTRO_RETRIES=1`), matching the iOS leg, and the script
  clears logcat itself before each attempt.

## Why this is over 500 lines

It is net deletion: 535 insertions against 1094 deletions. The bulk is the 901 dead lines plus 131
mechanical lines of the SQLDelight package move. Splitting it would mean landing exclusion filters
that reference a package that does not exist yet, or floors measured against a tree that still
contains the dead code.

## Testing

- `./gradlew testAndroidHostTest --continue` green across all modules (33 host-test tasks)
- `./gradlew koverVerify --rerun-tasks` green — forced a real run rather than trusting the cache,
  and confirmed no floor violations
- `./gradlew koverLog` reports merged line coverage 63.48% against the floor of 61
- `./gradlew detekt --continue` green, no autocorrections
- `actionlint` clean on all three touched workflows (only pre-existing SC2086 notes in untouched
  `run:` blocks)


ksharma-xyz:
* **#1912** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1912?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1912?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1911** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1911?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #1919, opened 2026-08-21
ci(maestro): hide emulator error dialogs so an ANR cannot eat every tap

ksharma-xyz:
## What

Sets `hide_error_dialogs` on the CI emulator so a system-app ANR dialog can no longer block the Maestro flows.

## Changes

- `maestro-pr-smoke.yml`, `maestro-nightly.yml`: `adb shell settings put global hide_error_dialogs 1` before running flows, alongside the existing animation-scale settings.

## Why

The smoke lane has been failing on most recent runs, both flows on their first assertion (`savedtrips.screen` never visible). The uploaded debug artifact shows the app healthy on the intro carousel with a "Pixel Launcher isn't responding" ANR dialog on top; the dialog is also present in the captured view hierarchy at the failing step. Taps aimed at `intro.action.primary` hit the dialog instead, so the carousel never advances and the home-screen wait times out. The launcher ANR depends on runner load, which made the lane look flaky rather than broken.

No signal is lost by hiding error dialogs: no flow asserts on them, real crashes are caught by the `FATAL EXCEPTION` logcat check in `run-flows.sh` (which reads the log, not the screen), and the runtime-permission dialogs driven by `06-permissions-denied` are a different dialog class and unaffected.

## Testing

- `actionlint` on both files: no new findings (pre-existing info-level SC2086 on untouched lines only).
- The smoke run on this PR itself is the real verification.

https://claude.ai/code/session_015CtZiJyHCYoNgULXRfPH9W

### DIFF
--- a/.github/workflows/maestro-nightly.yml
+++ b/.github/workflows/maestro-nightly.yml
@@ -97,13 +97,17 @@ jobs:
           # MAESTRO_RETRIES is a same-line prefix rather than an export — this
           # lane retries once, matching the iOS leg below. The script clears
           # logcat itself before each attempt.
-          # hide_error_dialogs: see maestro-pr-smoke.yml — a system app ANR
-          # ("Pixel Launcher isn't responding") otherwise parks a dialog over
-          # the app and steals every tap. Hides error dialogs only; the
-          # permission dialogs 06-permissions-denied drives are unaffected.
+          # hide_error_dialogs + launcher force-stop: see maestro-pr-smoke.yml —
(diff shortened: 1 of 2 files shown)


ksharma-xyz:
* **#1919** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1919?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1919?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Issue #1915, opened 2026-08-21
ModalBottomSheet loses its state on Activity recreation


ksharma-xyz:
Rotate (or theme-switch, or font-size change) with any ModalBottomSheet open and it closes. Cause: the sheet's Dialog fires onDismissRequest while being disposed during recreation, the caller's dismiss handler runs, and the host state is nulled as if the rider closed it.

Affects every sheet wired this way: service alerts, date-time selector, stop-details sheet, map options.

Notes from the investigation (fix/remember-saveable-ui-state work):
- rememberSaveable on the host state is necessary but not sufficient; the dismiss callback nulls it after restore
- StateRestorationTester cannot reproduce this (it never builds the dialog window); only device/emulator shows it
- Proven by re-tapping the same stop after a theme switch: one tap reopened the sheet, impossible if state had survived

Fix shape: one shared pattern that distinguishes user-initiated dismissal from disposal-during-recreation (e.g. only clear state when the sheet is dismissed while the lifecycle is RESUMED, or gate the callback on a saveable token), applied to all sheets, pinned by an emulator-verified check since host tests cannot see it. TimeTableEntry also has a dual-pane call-site hoist that was built and reverted because this defect masks its benefit; land it in the same pass.

ksharma-xyz:
## Investigation: the stated cause does not reproduce

Measured on an emulator (API 36) with a probe build that logged every `onDismissRequest` together with the lifecycle state and a stack trace:

- Rotating with a sheet open logged **no** `onDismissRequest` at all. The date-time selector sheet, whose visibility flag is already `rememberSaveable`, came back on screen after the rotation.
- The only `onDismissRequest` recorded in the whole session came from `ModalBottomSheet.kt:157` (`animateToDismiss`) at lifecycle `RESUMED`, i.e. a scrim tap.

So the dialog is not clearing the state during disposal. What actually closed sheets was ordinary lost state, plus one callback nobody suspected:

1. `SearchStopMap` and `JourneyMap` held the open stop in a plain `remember`.
2. `MapOptionsBottomSheet` held the rider's unsaved radius/mode/toggle edits in a plain `remember`, on a sheet that itself stays open, so the edits reverted under an unchanged-looking sheet.
3. `SearchStopScreen` switched from map back to list on any text-field callback. The field re-emits its restored text when the Activity is recreated, and that read as typing, so a configuration change took the rider off the map and the sheet with it.

[...]

PR #1922, opened 2026-08-21
fix(trip-planner): keep map stop sheets and pending map options across recreation

ksharma-xyz:
## What

Keeps the map stop-details sheets and the map options sheet's unsaved edits alive when the Activity is recreated, and stops the search screen dropping the rider back to the list on a configuration change.

## Changes

- `SearchStopMap.kt` and `JourneyMap.kt`: which stop's sheet is open is now a `rememberSaveable` stop id, resolved back to the feature from the map state the ViewModel still holds. That is the same lookup the map layer already does for a tap, so a stop that has left the displayed set leaves the sheet closed, which is the reset the old `remember(mapState)` key gave. Held in a plain `remember`, the sheet closed itself on every dark/light switch and font-size change.
- `MapOptionsBottomSheet.kt`: pending radius, mode set and toggles are saveable, with a `Saver` storing the mode set as its list of product-class ids. The sheet already stayed open across recreation, so the rider's unsaved edits reverted under a sheet that looked untouched.
- `SearchStopScreen.kt`: the map-to-list switch now ignores a text callback carrying unchanged text. The field re-emits its restored value when the Activity is recreated, and taking that for typing put the rider on the list on every configuration change, taking any open sheet with it. The comment above this call already recorded the same trap for focus callbacks. Extracted as `searchTextChangeHandler` so `SearchStopScreenSinglePane` stays under the cyclomatic limit; `rememberStopForSheet` extracted from `MapContent` for the same reason.
- `MapOptionsSheetRestoreTest`: new. Asserts what Save commits after recreation rather than which chip draws selected, since the committed value is what the rider loses. Verified to fail when the fix is reverted.
- `TimeTableStopSheetRestoreTest`: KDoc corrected. It stated as fact that `ModalBottomSheet` fires `onDismissRequest` while its dialog is disposed during recreation. A probe build logging every `onDismissRequest` with its lifecycle state recorded none at all across a rotation, and the sheets whose visibility flag was already saveable came back on screen. The only dismissals recorded were scrim taps at `RESUMED`.
- `docs/learning/2026-08-21-sheet-closed-by-a-text-callback.md`: new entry covering the wrong theory, the probe that settled it, and the `adb shell input tap` artefact where the injected `ACTION_UP` lands on the dialog window the `ACTION_DOWN` created and dismisses the sheet as an outside touch.

Left open, tracked in #1915: on a phone, rotating swaps single pane for dual pane, which use different map composables, so no state inside either can carry an open sheet across that switch. Closing it means hoisting the selection above `AdaptiveScreenContent` and giving both panes the same sheet, the way `showMap` already is.

## Testing

- `./gradlew :feature:trip-planner:ui:testAndroidHostTest` and `:taj:testAndroidHostTest` green
- `./gradlew detekt --continue` green, nothing left auto-corrected
- Emulator (Pixel-class, API 36), on the final build: stop sheet on the search map, stop sheet on the journey map and the options sheet with an unsaved 5km pick all survive a dark-mode recreation; journey map also rotated portrait to landscape and back with the sheet open; `adb logcat` shows no `FATAL EXCEPTION`
- Before the fix, the same steps closed the sheet and returned the search screen to the recents list

## Snapshots

No visual change: this PR changes which state survives recreation, not what any screen draws. No goldens moved.

Part of #1915

### DIFF
--- a/docs/learning/2026-08-21-sheet-closed-by-a-text-callback.md
+++ b/docs/learning/2026-08-21-sheet-closed-by-a-text-callback.md
@@ -1,0 +1,75 @@
+# A bottom sheet that "the dialog dismissed itself" was closed by a text field
+
+**2026-08-21** · Search-stop map, journey map, map options · ~4 builds, 3 install cycles, one evening
+
+## Symptom
+
+Open a stop's details sheet on the search-stop map, switch the device to dark mode (or change
+font size), and the sheet is gone. The screen behind it is gone too: the map has been replaced
+by the recents list, as if the rider had never tapped "Select on map".
+
+## Root cause
+
+Three separate faults, none of them the one that had been written down.
+
+1. `SearchStopMap` and `JourneyMap` held which stop's sheet is open in a plain `remember`, so
+   Activity recreation dropped it. `MapOptionsBottomSheet` did the same with the rider's
+   unsaved radius/mode/toggle edits, on a sheet that itself stays open — so the edits reverted
+   under an unchanged-looking sheet.
+2. `SearchStopScreen` switched from map back to list on any text-field callback. The field
+   re-emits its restored value when the Activity is recreated, and that read as typing. The
+   comment right above it already recorded the same trap for focus callbacks; the value
+   callback had the same problem and was not covered.
+3. Phone portrait and landscape are different layouts (single pane vs dual pane) and use two
+   different map composables, so no `rememberSaveable` inside either can carry state across a
+   rotation. That one is still open — see the gap below.
+
+## Why it took so long
+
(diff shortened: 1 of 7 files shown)


ksharma-xyz:
* **#1922** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1922?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1922?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1933, opened 2026-08-21
fix(ai-text): report one availability vocabulary from both platforms

ksharma-xyz:
## What

Both platform AI services reported unavailability in their own words, and both callers only understood the Android ones. Adds a shared vocabulary and maps each platform onto it.

## Changes

- `AiUnavailableReasons` in `core/ai-text` commonMain: `MODEL_DOWNLOADING`, `DEVICE_UNSUPPORTED`, `NEEDS_DEVICE_SETTING`, `CHECK_FAILED`. Modelled on the existing `SpeechUnavailableReasons`, which was added for the same class of bug.
- `AndroidAiTextService` maps ML Kit's `FeatureStatus` onto it. The raw status integer moves out of the reason string and into the log line, since the reason is a value callers compare against and a glued-on number broke every comparison.
- `AiTextBridge.swift` maps `SystemLanguageModel.Availability.UnavailableReason` cases explicitly instead of interpolating the enum into a string.
- `AlertSummaryViewModel.RETRIABLE_REASONS` held the same two raw ML Kit strings and had the same hole, so it reads the shared constants now.

Note that `modelNotReady` on iOS means the model is still downloading, which is retriable. Under the old string comparison it was classified alongside a device that can never run the model.

The `summarize` path keeps its silent-degradation contract; only the reason strings changed there.

## Testing

- `./gradlew testAndroidHostTest --continue` green, including `AlertSummaryViewModelTest` updated to the shared constants
- Verified on a physical Pixel 10 Pro: `checkExtractionAvailability -> Available`. On an emulator with no Gemini Nano, `FeatureStatus` 0 maps to `device_unsupported` as intended.

The Swift changes are compile-verified through the Kotlin/Native bridge only; Foundation Models availability branches cannot run on the simulator, which is an Apple platform restriction.

### DIFF
--- a/core/ai-text/src/androidMain/kotlin/xyz/ksharma/krail/core/aitext/AndroidAiTextService.kt
+++ b/core/ai-text/src/androidMain/kotlin/xyz/ksharma/krail/core/aitext/AndroidAiTextService.kt
@@ -17,6 +17,9 @@ import kotlinx.coroutines.flow.catch
 import kotlinx.coroutines.guava.await
 import kotlinx.coroutines.launch
 import kotlinx.serialization.json.Json
+import xyz.ksharma.krail.core.aitext.AiUnavailableReasons.CHECK_FAILED
+import xyz.ksharma.krail.core.aitext.AiUnavailableReasons.DEVICE_UNSUPPORTED
+import xyz.ksharma.krail.core.aitext.AiUnavailableReasons.MODEL_DOWNLOADING
 import xyz.ksharma.krail.core.log.log
 
 /**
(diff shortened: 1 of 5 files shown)


ksharma-xyz:
* **#1933** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1933?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1933?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1934, opened 2026-08-21
fix(trip-planner): tell the rider why Ask KRAIL could not read their trip

ksharma-xyz:
## What

Every model-unavailable path landed in `UNRESOLVED` with no reason set, which the banner renders as "Something is missing there. Name where you are going, and where from." That is advice about a sentence, shown to a rider whose sentence was never read.

## Changes

- Adds `UnresolvedReason.MODEL_UNAVAILABLE` and `MODEL_NEEDS_SETTING`, with messages naming what the rider can act on. On a device that cannot run the model, the message points at tapping the stops instead; on one with the setting switched off, it names the setting.
- `withUnavailableModel` in a new `TripIntentResolution.kt` maps each `AiUnavailableReasons` value onto a phase and reason. `CHECK_FAILED` falls through to `COULD_NOT_READ`, since a failed check says nothing about the device and is worth retrying. It is matched by fallthrough rather than equality because Android suffixes the throwable message onto it.
- Adds `isDeviceCapable` to the state and `isWayInAvailable` (flag AND device) as the entry-point gate. The wheel was gated on the remote-config flag alone, so it rendered on devices with no on-device AI and failed on every submit.
- Only `DEVICE_UNSUPPORTED` removes the entry point. A model still downloading, or one switched off in Settings, both keep it, because both become available without a new build.
- `isDeviceCapable` is held as a ViewModel field as well as in the state, since `OpenInput`, `StartOver` and `CloseInput` each rebuild the state from scratch.
- Pure functions live in a separate file rather than as private ViewModel methods: detekt counts branches in nested functions toward the enclosing one, so the extraction only helps if it leaves the class.

The existing `unsupported device lands in UNRESOLVED` test asserted the phase and never the message, which is why this was green while the wrong text shipped. New tests pin the reason.

## Testing

- `./gradlew testAndroidHostTest --continue` green; 6 new tests covering each reason, the entry-point gate, and the device verdict surviving all three state resets
- Physical Pixel 10 Pro: `[AI_OUTCOME] deviceCapable=true`, entry point present, a submit resolved both ends. Emulator without Gemini Nano: `deviceCapable=false`, entry point correctly absent.

## Snapshots

Not captured. The banner copy only renders on a device where the model is unavailable, and the emulator that reproduces that state also hides the entry point needed to reach the banner. Verified through unit tests on the reason mapping instead.

### DIFF
--- a/feature/trip-planner/ui/src/commonMain/kotlin/xyz/ksharma/krail/trip/planner/ui/components/ai/AiInputPieces.kt
+++ b/feature/trip-planner/ui/src/commonMain/kotlin/xyz/ksharma/krail/trip/planner/ui/components/ai/AiInputPieces.kt
@@ -275,6 +275,17 @@ private fun AiSearchInputUiState.unresolvedMessage(): String = when (unresolvedR
     UnresolvedReason.COULD_NOT_READ ->
         "That did not come through. Have another go, or use fewer words."
 
+    // Neither the sentence nor anything the rider can do. Saying so and pointing at the way
+    // that does work beats an apology, and beats the advice this used to give, which was to
+    // reword a sentence nothing had read.
+    UnresolvedReason.MODEL_UNAVAILABLE ->
+        "This phone cannot run the on device AI that Ask KRAIL needs. Tap the stops to plan a " +
+            "trip instead."
+
+    // The one model problem with a fix the rider owns, so it names the switch.
+    UnresolvedReason.MODEL_NEEDS_SETTING ->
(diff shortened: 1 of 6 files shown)


ksharma-xyz:
* **#1934** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1934?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1934?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1935, opened 2026-08-21
fix(trip-planner): never resolve a trip from a stop to itself

ksharma-xyz:
## What

"lets go to work" came back from the iOS model with "work" in both `originText` and `destinationText`. Both resolved to the rider's Work stop and the home row was filled Work to Work.

## Changes

- `resolveTripOrigin` drops an origin that resolved to the destination and falls through to the nearby-stop fallback an unstated origin already takes, which knows to exclude it. If that also comes back empty the field is left blank and editable.
- `sameStopAs` compares stop id **or** name. One station carries several ids in this data (a label pinned to the parent, a search result on a platform), so an id check alone let the same station through twice under two ids. The name comparison is safe here because the only decision it drives is dropping an origin the rider never asked for.
- The nearby-stop fallback result is re-checked by the same predicate, for the same reason.
- A named origin that resolved to nothing keeps its text on the row as before. Only one equal to the destination is dropped, and the destination is always the field kept, since that is the end the rider actually named.
- Brings the iOS extraction prompt back in step with the Android one: adds the worked examples, adds "let's go to X" and "lets go to X" to the lone-place rule, and states that one place must never fill both fields.

`withSinglePlaceInTheRightField` could not catch this: it only fires when exactly one field is filled, and here both were.

The prompt change alone would not be sufficient. A prompt is advice, and this repo already learned that from `SinglePlaceIntent.kt`, so the guard is deterministic and tested.

## Testing

- `./gradlew testAndroidHostTest --continue` green; 3 new tests covering the same place in both fields, two ids for one station, and a genuinely different origin being left alone
- Physical Pixel 10 Pro: a real submit resolved two distinct stops, confirming the guard does not over-fire on a legitimate two-ended trip

### DIFF
--- a/core/ai-text/src/swift/aiTextBridge/AiTextBridge.swift
+++ b/core/ai-text/src/swift/aiTextBridge/AiTextBridge.swift
@@ -74,20 +74,48 @@ import FoundationModels
         #if canImport(FoundationModels)
         if #available(iOS 26.0, *) {
             Task {
+                // Kept deliberately in step with `buildExtractionPrompt` in
+                // AndroidAiTextService. This side used to be a shortened paraphrase of it
+                // with no examples, and the two drifted where it mattered: Android listed
+                // "let's go to X" in the lone-place rule and carried three worked examples,
+                // this one listed neither. "lets go to work" came back with work in BOTH
+                // fields and the rider got a trip from their work stop to their work stop.
+                // The deterministic guard in AiSearchInputViewModel is what actually stops
+                // that reaching the row; this is the half that stops the model producing it.
                 let instructions = """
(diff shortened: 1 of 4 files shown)


ksharma-xyz:
* **#1935** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1935?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1935?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1936, opened 2026-08-21
fix(trip-planner): let riders dismiss the Ask KRAIL dialog

ksharma-xyz:
## What

Tapping outside the Ask KRAIL dialog did nothing, on either platform. On iOS that left the card with no way out at all.

## Changes

- The scrim handles the dismiss tap itself, with the card swallowing its own via `detectTapGestures` so a tap on the card does not close it.
- iOS additionally gets a close button in the dialog header. Android does not draw one: the system back gesture is already a way out, and a second control is clutter on a card this size. Gated with the existing `getAppPlatformType()` pattern.
- Adds `CloseIcon` to taj, stroked at the same weight as `SendIcon` and `MicIcon`.
- The title's horizontal clearance is only claimed where the button is drawn, so it stays centred on Android.
- The full-screen variant used past `ACTIONS_STACK_SCALE` already had a title bar back action and is untouched.

### Why the platform default did not work

Compose Multiplatform decides "outside" geometrically. In `Dialog.skiko.kt`:

```kotlin
layer.setOutsidePointerEventListener(onOutsidePointerEvent)
...
layer.boundsInWindow = IntRect(positionWithInsets, contentSize)
```

`contentSize` is the measured size of the dialog's content. This content fills the window, which it must in order to centre the card and own the keyboard inset, so the bounds were the whole screen and no pointer was ever outside them. Android survived on `dismissOnBackPress`; iOS has neither a back press nor a swipe on a Compose dialog.

`ModalBottomSheet.ios.kt` already solves this the same way for its own scrim, so this brings the dialog in line with a pattern the codebase had.

This is the only Compose `Dialog` in the app, so the trap is not repeated elsewhere.

## Testing

- `./gradlew testAndroidHostTest --continue` green
- `./gradlew detekt --continue` green
- `compileDebugSources` and `compileKotlinIosSimulatorArm64` green
- Installed on a physical Pixel 10 Pro, where the dialog is reachable

## Snapshots

Not captured. The dialog only renders where the on-device model is available, and the one device here that satisfies that is a personal phone, so no screenshots of it are attached. Worth a look on device before approving, since the header layout is the visible part of this change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01M8SAxiB2G4nGs66NBvNkVb

PR #1937, opened 2026-08-21
fix(speech-to-text): end an iOS listening session when the rider stops

ksharma-xyz:
## What

An iOS listening session only ended when the ViewModel's ten to fifteen second ceiling fired. Android's ends after about four seconds of quiet, so the same spoken sentence took far longer to finish on iOS. Also fixes two other iOS-only faults in the same file.

### Ending the session

`SFSpeechAudioBufferRecognitionRequest` never decides on its own that the rider has finished: `isFinal` arrives only after `endAudio()`, and nothing called it. The ViewModel ceiling exists as a backstop for a recogniser that never reports an end, and on iOS it had quietly become the only mechanism.

- `watchForSilence` ends the audio once the transcript stops changing: four seconds mid sentence, six before any word is recognised. The windows are matched to `SILENCE_THAT_ENDS_THE_SESSION_MILLIS` on the Android side, kept in step by hand since the two live in different source sets.
- It watches the transcript, not the audio level, because the tap sees room noise on every buffer while the transcript only moves when words are recognised.
- `endAudio()` rather than closing the flow, so the recogniser still delivers its final transcript. Same graceful finish `stopListening` asks for.

### Audio session

The session was switched to the record category and left there, with no deactivation carrying `NotifyOthersOnDeactivation`, so a rider who used the mic once lost whatever they had playing until the app was killed. It is now activated explicitly and released on close.

An older comment claimed cinterop does not expose `setActive` at any arity. That is not true on this Kotlin version; both forms resolve, and the comment is removed rather than left to mislead.

### Recogniser construction

`SFSpeechRecognizer()` returns nil when the device locale has no speech model. Cinterop declares that initialiser non-null, so the nil arrived as a thrown error, from an eagerly initialised field on a Koin `single`, during ViewModel construction on home-screen entry. That is a crash of the app rather than of the feature.

It is now lazy, nullable, and built inside `runCatching`, falling back to the default-locale recogniser, so an unusable one reports `NOT_AVAILABLE` like any other dead end.

## Testing

- `./gradlew testAndroidHostTest --continue` green (no Android behaviour changes here; the Android service is untouched)
- `compileKotlinIosSimulatorArm64` and `compileDebugSources` green

Not verified at runtime. The changes are iOS-only and speech recognition needs a physical iOS device, which was not available for this branch. The silence windows in particular are worth confirming by speaking a sentence with a pause in it before this is relied on.

### DIFF
--- a/core/speech-to-text/src/iosMain/kotlin/xyz/ksharma/krail/core/speechtotext/IosSpeechToTextService.kt
+++ b/core/speech-to-text/src/iosMain/kotlin/xyz/ksharma/krail/core/speechtotext/IosSpeechToTextService.kt
@@ -2,19 +2,28 @@ package xyz.ksharma.krail.core.speechtotext
 
 import kotlinx.cinterop.ExperimentalForeignApi
 import kotlinx.coroutines.channels.awaitClose
+import kotlinx.coroutines.currentCoroutineContext
+import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.callbackFlow
+import kotlinx.coroutines.isActive
+import kotlinx.coroutines.launch
 import kotlinx.coroutines.suspendCancellableCoroutine
 import platform.AVFAudio.AVAudioApplication
 import platform.AVFAudio.AVAudioApplicationRecordPermissionGranted
 import platform.AVFAudio.AVAudioEngine
 import platform.AVFAudio.AVAudioSession
 import platform.AVFAudio.AVAudioSessionCategoryRecord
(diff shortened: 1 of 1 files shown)


ksharma-xyz:
* **#1937** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1937?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1937?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1940, opened 2026-08-22
fix(taj): stop the AI gradient ring leaving a pale rim at its corners

ksharma-xyz:
## What

The AI gradient ring left a pale rim at its corners. Reported on the Ask KRAIL dialog, where it reads as a white border between the gradient and the card.

## The bug

The ring is stroked along a rect inset by half the stroke width, but the centreline was given the **surface's** corner radius:

```kotlin
val half = strokePx / 2f
drawRoundRect(
    style = Stroke(width = strokePx),
    cornerRadius = CornerRadius(cornerRadius.toPx()),   // full radius on the centreline
    topLeft = Offset(half, half),
    size = Size(size.width - strokePx, size.height - strokePx),
)
```

A stroke's outer edge curves at the centreline radius plus half the width, so the outer edge curved at `R + half` while the card's `clip` curves at `R`. A tighter-cornered ring than the card it was tracing: at every corner it pulled inward and the background showed through outside it.

Straight edges were fine, which is why this looked like a rendering oddity rather than a geometry error.

## Measured

Walking the top-left diagonal inward on the recorded snapshot, classifying each pixel as backdrop / gradient / white:

```
before   backdrop(43)  gradient(1)  WHITE(1)  gradient(12)  WHITE(83)
after    backdrop(43)  gradient(12)           WHITE(85)
```

The ring was split by a band of card background.

## Changes

- The centreline now uses `cornerRadius - half`, so the stroke's outer edge lands exactly on the shape's own radius. Floored at zero, so a stroke thicker than the radius degrades to square corners rather than inverting into a negative radius.
- Adds a snapshot preview reproducing the failure conditions: light surface, this ring, dark backdrop.

The fix is in the modifier, so all three callers get it: the Ask KRAIL dialog, `AiInputBar`, and `AiAlertSummaryCard`.

## Why it was not caught

Every existing preview of this modifier puts the ring on a surface the same colour as the gap. With nothing behind it to contrast against, there is nothing to see. The new preview is deliberately a light card on a dark backdrop for that reason.

## Testing

- Snapshots recorded and the committed golden re-checked pixel by pixel; the diagonal above is read off the file in the PR

## Snapshots

| Case | Result |
|---|---|
| Ring on a light surface, dark backdrop |  |

### DIFF
--- a/taj/screenshots/PreviewAiGradientBorderOnFilledSurface.1._Light_Mode_Screen_WITH_BACKGROUND_WIDTH_411DP_HEIGHT_891DP_dark_normal.png
+++ b/taj/screenshots/PreviewAiGradientBorderOnFilledSurface.1._Light_Mode_Screen_WITH_BACKGROUND_WIDTH_411DP_HEIGHT_891DP_dark_normal.png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:6cfff8ba7d380ec3e1fb83ff7d949d86cca2c039b1d9c7994b05236540dbfa30
+size 46498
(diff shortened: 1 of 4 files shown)


ksharma-xyz:
* **#1940** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1940?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1940?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1941, opened 2026-08-22
feat(taj): add a contrast helper that can reach any target

ksharma-xyz:
## What

Adds `Color.ensureContrastWith`, a contrast-adaptation helper that can reach its target for any
colour on any ground. Nothing calls it yet; `ensureMinimumContrast` is untouched.

## Changes

- `A11yColors.kt`: new `ensureContrastWith(background, toward, minContrast)`. It blends toward the
  ground's own content colour rather than raising HSV value, so the ladder always ends on a colour
  that clears the threshold by definition.
- Walks a 32-rung ladder from the smallest blend up rather than binary searching. Contrast is not
  monotone along the blend because Compose's `lerp` interpolates in Oklab, and 110 of 7680 measured
  ladders reverse somewhere in the middle. A scan needs no monotonicity and returns the
  least-adapted rung that clears.
- Quantising to rungs keeps output stable: a one-digit edit to a surface token cannot shift every
  derived colour by 1/255 and churn screenshot goldens.
- `CONTRAST_SAFETY_MARGIN` of 0.05 so a result never lands exactly on the threshold, where float
  comparison and 8-bit quantisation can push it under.
- `ContrastAdaptationTest`: sweeps 1920 hue/saturation/value combinations against every opaque
  ground in both schemes at both thresholds and asserts every one reaches its target. Includes a
  scan-broke detector so it cannot pass vacuously, and pins the old helper's failure so the
  justification for the new one cannot rot.

Why the existing helper was not enough: `ensureMinimumContrast` raises HSV value only, so a colour
already at full value cannot be adapted further, and when the loop runs out of steps it returns the
colour anyway with no signal. Over the same sweep it returns a still-failing colour for 24% of
combinations against the bottom-sheet ground and 14% against the dark surface. Light mode is
unaffected because darkening has no floor, which is why the gap was not visible.

## Testing

- `./gradlew :taj:testAndroidHostTest` green

### DIFF
--- a/taj/src/commonMain/kotlin/xyz/ksharma/krail/taj/theme/A11yColors.kt
+++ b/taj/src/commonMain/kotlin/xyz/ksharma/krail/taj/theme/A11yColors.kt
@@ -1,6 +1,7 @@
 package xyz.ksharma.krail.taj.theme
 
 import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.lerp
 import androidx.compose.ui.graphics.luminance
 import xyz.ksharma.krail.taj.brighten
 import xyz.ksharma.krail.taj.darken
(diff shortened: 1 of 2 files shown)


ksharma-xyz:
* **#1945** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1945?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1944** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1944?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1943** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1943?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1942** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1942?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1941** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1941?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1941?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1943, opened 2026-08-22
fix(taj): read theme colour as ink where it is drawn on a surface

ksharma-xyz:
## What

Moves every call site onto the role accessor that matches what it draws. The colour changes only
where the theme colour was being read as ink against an app surface.

## Changes

Ink, now adapted:

- `ButtonDefaults.textButtonColors` is the widest one. A text button has a transparent container, so
  its label sits directly on the surface. Covers "Not now", "Reset" and "Show previous departures".
- `TextFieldPlaceholderDefaults.colors` activeColor, which is the chosen stop name on the search row.
- The KRAIL wordmark, the app upgrade prompt, the Park and Ride add ring.
- Held to 3.0 rather than 4.5 because they are strokes and icons: the origin/destination timeline,
  loading dots, card borders, the pager indicator, the unselected Discover chip border, date chevrons.
- The time picker's hand and period border, adapted against the dial via `themeInkColorOn` rather
  than the surface. The dial is filled with the theme colour at 45%, so the hand was purple on
  purple. Every theme failed that pairing, not only Purple Drip.

The dial is composited over the sheet before being measured against. `contrastRatio` reads
`luminance()`, which ignores alpha, so passing the translucent wash straight in reads it as far
lighter than it renders and over-adapts the hand to near-white.

`OutlineRadioButton` carried both roles in one `themeColor` parameter: selected fills with it,
unselected draws a label and border with it. Split into `groundColor` and `inkColor` so the selected
chip keeps the brand colour and the unselected one becomes legible. Its label is text and takes the
4.5 ink; the date chevrons beside it are icons and take 3.0.

Left unchanged: ground and decoration keep the raw colour and only change accessor. Card fills, the
`P` badge, the map disc, the Discover chip fill, ripples and the cloud gradient.

No remaining callers of the deprecated `themeColor()`.

## Testing

- `./gradlew testAndroidHostTest --continue` green across all modules
- `./gradlew detekt --continue` green
- `compileDebugSources` and `compileKotlinIosSimulatorArm64` green
- Device QA on an emulator, Purple Drip theme: opened the home screen, search stop screen, timetable
  and the plan-your-trip sheet; rotated to landscape and back on each with data loaded; switched
  light and dark; navigated away and back. No `FATAL EXCEPTION` in logcat.

Measured from device screenshots, Purple Drip in dark mode:

| Element | Before | After | Needs |
|---|---|---|---|
| Chosen stop name on the search row | 2.95 | 6.11 | 4.5 |
[...]

PR #1942, opened 2026-08-22
feat(taj): split the theme colour into ground, ink, wash and decoration

ksharma-xyz:
## What

Splits `themeColor()` into four role accessors so each call site says which contrast obligation it
carries, and derives the ink role so it is legible on every ground. No call sites change yet.

## Changes

- `ColorsExt.kt`: adds `themeGroundColor`, `themeInkColor`, `themeInkColorOn` and `themeDecorColor`
  beside the existing `themeBackgroundColor` wash. Deprecates `themeColor`.
- `ThemeInk.kt`: `asThemeInk` derives the ink once per theme and scheme against the hardest ground
  in that scheme. The hardest ground is picked by luminance from `inkGrounds()` rather than named,
  so adding a surface token cannot silently invalidate every ink.
- `ThemeInkContrastTest`: measures every `KrailThemeStyle` against every ground in both schemes at
  both thresholds, self-healing over the enum and the ground list. Also holds `inkGrounds()`
  against `KrailColors`, so a new ground token has to be listed or excused with a reason. No
  known-failure set, and there should never be one.

Only ink is adapted. Ground keeps the raw colour because content drawn on it derives its own colour
from `getForegroundColor`, which `ThemeContrastTest` already covers. Decoration carries no
obligation at all, and saying so explicitly is what lets a future rule on ink stay unconditional
rather than needing a permanent baseline.

Why one ink per scheme rather than one per ground: adapting against whatever ground the ink lands on
produces a different shade on the surface, on a sheet and on a card, and a sheet opened over a
screen shows two of them at once. Deriving against the hardest ground means one value clears them
all. `themeInkColorOn` exists for the genuine exception, ink on a ground that is not an app surface.

Why this gap existed: `ThemeContrastTest` measures content drawn on a theme colour,
`KrailColorTokenContrastTest` measures the fixed tokens, and `TransportModeContrastTest` iterates
`TransportMode.all`, which does not include the theme-only colours. Nothing measured a theme colour
used as ink against the surfaces it is drawn on.

## Testing

- `./gradlew :taj:testAndroidHostTest` green
- `./gradlew detekt --continue` green
- `compileDebugSources` and `compileKotlinIosSimulatorArm64` green

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1944, opened 2026-08-22
refactor(taj): declare each theme's AI gradient partner on the enum

ksharma-xyz:
## What

Moves each theme's AI gradient partner from a lookup map with a default onto a `KrailThemeStyle`
constructor parameter, so a new theme cannot be added without choosing one.

## Changes

- `KrailThemeStyle` gains `aiGradientPartner: Color`. The per-pair reasoning that was in
  `AiThemeGradientTokens`' map comments moves with it.
- `AiThemeGradientTokens.stopsFor(style)` reads the partner off the enum. The `partners` map and its
  `?: partners.getValue(DEFAULT_THEME_STYLE)` both go.
- `ThemeGradientPairTest`: measures the hue distance for every entry, asserts a theme never partners
  itself, and asserts the middle stop stays distinct from both ends. Self-healing over the enum.
- Drops the now-stale `AiThemeGradientTokens.kt` entry from `BrandHexUniquenessTest`'s allowlist.
  Removing the mode-colour literals from that file left the entry excusing something that no longer
  exists, which the guard's own stale check flagged.

Why this shape: a theme missing from the old map did not fail to build and did not draw nothing, it
silently wore Train's gradient, and nothing would have caught it. A constructor parameter makes the
same mistake a compile error at the line being edited, rather than a test failure in another file or
a wrong colour on a device.

A constructor parameter can only force a choice, not a good one, hence the hue-distance test.
Writing it turned up that the invariant stated in the KDoc was wrong: it claimed every pair sits
between 97 and 134 degrees apart, but Bus into Magic Yellow is 147.5 and Purple Drip into Bus is
96.8. The window is now set from the design rule the KDoc actually argues for, and the prose is
corrected to the measured numbers.

Left unchanged: `stopsFor(hex)` keeps its fallback. An unknown hex can only come from preferences
written by an older build, so falling back is correct there, and a test pins that it still yields a
usable gradient.

## Testing

- `./gradlew testAndroidHostTest --continue` green across all modules

No rendered output changes: every shipped theme keeps the partner it already had.

### DIFF
--- a/feature/trip-planner/ui/src/androidHostTest/kotlin/xyz/ksharma/krail/trip/planner/ui/contrast/BrandHexUniquenessTest.kt
+++ b/feature/trip-planner/ui/src/androidHostTest/kotlin/xyz/ksharma/krail/trip/planner/ui/contrast/BrandHexUniquenessTest.kt
@@ -106,8 +106,6 @@ class BrandHexUniquenessTest {
                 "taj cannot depend on :core:transport; these are the theme tokens themselves.",
             "taj/src/commonMain/kotlin/xyz/ksharma/krail/taj/tokens/AiGradientTokens.kt" to
                 "Same: the AI gradient's stops are mode colours, declared taj-side.",
-            "taj/src/commonMain/kotlin/xyz/ksharma/krail/taj/tokens/AiThemeGradientTokens.kt" to
-                "Same, for the per-theme variant of that gradient.",
             // Preview and sample fixtures. Each should take the colour from TransportMode.
             "feature/trip-planner/ui/src/commonMain/kotlin/xyz/ksharma/krail/trip/planner/ui/" +
                 "components/DepartureBoardStopCard.kt" to "Preview fixture.",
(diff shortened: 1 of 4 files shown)


ksharma-xyz:
* **#1945** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1945?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1944** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1944?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1944?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1943** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1943?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1942** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1942?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1941** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1941?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #1938, opened 2026-08-21
ci(maestro): stop running the smoke lane on pull requests

ksharma-xyz:
## What

Deletes the Maestro PR smoke workflow. The nightly lane already runs the same flows, so no coverage is lost.

## Changes

- Removes `.github/workflows/maestro-pr-smoke.yml`
- Inlines the two comments in `maestro-nightly.yml` that referenced the deleted file, rather than leaving them pointing at nothing
- Updates the lanes table and surrounding prose in `docs/testing/LAYERS.md`, including the line that claimed smoke gates a merge

## Why

The lane booted a full emulator on every non-draft PR and had been failing on all of them since #1920, regardless of what the PR changed. It was never a required status check (`main`'s ruleset requires only the three build jobs and `code-quality / detekt`), so it gated nothing. A permanently red mark that blocks nothing is worse than no mark, because people learn to scroll past it.

`.maestro/nightly/` runs the whole tree, smoke flows included, on Android and on an iPhone simulator.

## What this does not fix

The underlying failure. `02-plan-trip` taps `searchrow.from` and then times out waiting for `searchstop.query`, so the tap is not reaching the search screen on a software-rendered emulator:

```
[Failed] 02-plan-trip (44s) (Assertion is false: id: searchstop.query is visible)
```

The nightly lane fails the same way and stays red after this. That is a real signal and it should be chased on the nightly, where it is a report rather than a merge blocker. There are already two commits in this repo fighting emulator ANR dialogs stealing taps (`c49740d54`, `f08094fd3`), so that is the likely neighbourhood.

Worth being explicit: after this merges, nothing in CI gates a merge on a Maestro flow. An end-to-end regression is caught by reading the nightly.

## Testing

- `actionlint` on `maestro-nightly.yml`: 8 findings before and after this change, all pre-existing `SC2086` notes in untouched steps
- The nightly workflow still parses as YAML
- No source or test changes, so no Gradle tasks are affected

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01M8SAxiB2G4nGs66NBvNkVb

PR #1946, opened 2026-08-22
fix(maestro): make the nightly lane actually run its flows

ksharma-xyz:
## What

The nightly Maestro lane has never run a single flow. This fixes that, and sizes the one flow that then turned out to be genuinely flaky.

## The real bug

The lane passes `.maestro/` so it can cover the whole tree. Maestro does not recurse: pointed at a directory whose flows all live one level down, it finds nothing and exits non-zero with

```
Top-level directories do not contain any Flows: .../.maestro
```

Because that still exits 1, the lane looked like a red **test suite** rather than one that never started. Reproduced locally against the same commit, same message.

`.maestro/config.yaml` names the two lane directories. `shared/` is deliberately excluded (helper flows invoked via `runFlow`; listing it would run each as a test in its own right, which is what keeping them outside both lanes was for), and `ci/` too, since it holds the shell runner.

### Correcting the previous PR

#1938 claimed the nightly failed the same way as the PR smoke lane, on the `02-plan-trip` assertion. That was wrong. The smoke lane passes `.maestro/smoke/`, a directory with flows directly in it, so it always found them. The nightly's failure was never the same one.

## Supporting changes

**The settle wait now looks for a plateau, not a threshold.** #1939 waited for CPU under 85%. It ran, and disproved its own premise: the device reported a flat 100% on every sample for the full 120s, so the wait was pure dead time. An absolute threshold only means something if the idle figure is known, and here it is a property of the runner. Waiting for the reading to stop *moving* works whether the plateau sits at 30% or 100%, and now settles in ~5s.

**`google_atd` instead of `google_apis`.** ATD is Google's CI image: Play Services kept for Firebase, the rest of the app suite stripped. Honest caveat: this did **not** measurably move the reported CPU figure, which still reads 99%. It is kept because the apps it removes were the ones competing for the two cores the emulator gets, and a single aggregate sample is weak evidence either way. If it is judged unproven, it is the one change here safe to drop.

**`02-plan-trip` timeouts.** Once the lane ran, 5 of 6 flows passed and only this one failed, at a *different step on each attempt of the same run*: first the search screen at 15s, then on the retry all the way to the timetable and the journey list at 45s. Two different steps timing out in one run is a budget too tight, not a broken screen. Screen transitions go to 30s; the journey wait to 90s, being the only one with a live NSW API round trip on top of the render. The reasoning is written into the flow so the numbers do not read as arbitrary.

## Testing

Local, against the merged `main`:
- `maestro test .maestro/` before this: instant `Top-level directories do not contain any Flows`
- after: discovers and runs both lanes, whole suite **exit code 0**
- the plateau settle block run standalone against an emulator: `Device settled: CPU 6% (was 6%)`

CI, dispatched on this branch:
- before the config fix: no flows, immediate failure
- after: `01-launch-home`, `03-rotation-sweep`, `04-background-foreground`, `05-kill-relaunch`, `06-permissions-denied` all **Passed**; `02-plan-trip` failed on timing, which is what the last commit addresses
- a run with the raised timeouts is queued at time of writing

## Note on gating

This lane still gates nothing, deliberately. It should be watched for a stretch of consistently green runs before anyone considers making it a required check.

### DIFF
--- a/.github/workflows/maestro-nightly.yml
+++ b/.github/workflows/maestro-nightly.yml
@@ -86,7 +86,15 @@ jobs:
         uses: reactivecircus/android-emulator-runner@v2
         with:
           api-level: 34
-          target: google_apis
+          # google_atd, not google_apis. The ATD ("Automated Test Device") images
+          # are Google's own CI build: Play Services kept, because Firebase needs
+          # them, and the rest of the Google app suite stripped out. On the
+          # google_apis image the emulator sat at 100% CPU for the entire two
+          # minutes the runner waited for it, with TTS, YouTube Music, the
+          # launcher and quicksearch between them taking more of the machine than
+          # anything under test. That is not a boot storm that passes, which is
+          # what the previous attempt at this assumed; it is the image.
+          target: google_atd
           arch: x86_64
           profile: pixel_6
           disable-animations: true
--- a/.maestro/ci/run-flows.sh
+++ b/.maestro/ci/run-flows.sh
@@ -95,14 +95,23 @@ done
 
 # `dumpsys cpuinfo` prints the same "N% TOTAL" line the ANR dump above did, over
 # a window since the previous call, and unlike /proc/pressure it is readable by
-# the shell user (SELinux denies that one, which is why this does not use the
-# figure the dump quoted).
+# the shell user (SELinux denies that one).
 #
-# 85 rather than something tighter: an idle emulator still reads around 60% here,
-# because software rendering and the emulated kernel are themselves the load. The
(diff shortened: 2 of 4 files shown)


ksharma-xyz:
* **#1946** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1946?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1946?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1949, opened 2026-08-22
fix(speech-to-text): recognise on device on iOS, as Android already does

ksharma-xyz:
## What

Two bugs, found together, because the second hid the first.

## 1. No test tag inside any bottom sheet was ever visible to UI automation

The app calls `exposeTestTagsToUiAutomation()` once, on `KrailNavHost`. A `ModalBottomSheet` does not render inside that subtree — it gets its own window — so `testTagsAsResourceId` never reached it.

Nothing failed loudly. A selector for something in a sheet simply never matched, which reads as the element not being on screen. Found by watching a flow fail to tap a button plainly visible in the hierarchy dump, sitting there with an empty `resource-id` beside its own label:

```
id='' text='Save' acc=''
```

Fixed on taj's Android `ModalBottomSheet`, so every sheet in the app gets it, not just this one.

## 2. 02-plan-trip could not reach the search field

`SearchStopViewModel` opens the map options sheet by itself the first time a rider reaches stop search, gated on `KEY_HAS_SEEN_MAP_OPTIONS_SHEET`. While up, it covers `searchstop.query`.

On a fresh device that field is not slow to appear, it is **unreachable**. That is why doubling the timeouts in #1946 changed nothing, and why the flow passes on a developer emulator that dismissed the sheet months ago. The screenshot Maestro captured at the failure is a full-screen map with the sheet over it; the hierarchy confirms it:

```
'Close sheet' / 'Drag handle' / 'Map Options'
'Tap any stop on the map to select it — no typing needed.'
'Search Distance' / 1km 3km 5km / ...
```

## Changes

- `testTagsAsResourceId` on taj's Android `ModalBottomSheet`
- `SEARCH_STOP_MAP_OPTIONS_DONE` tag on the sheet's Done button
- `.maestro/shared/dismiss-map-options.yaml`, same shape as `dismiss-intro.yaml`: unconditional at the call site, `optional` inside, so a device that has already seen the sheet pays one timeout and moves on
- Both stop pickers call it, since either can be the first visit

It taps **Done** rather than the scrim, because Done is what writes `KEY_HAS_SEEN_MAP_OPTIONS_SHEET`. Dismissing any other way leaves the flag unset and the sheet returns for the destination half of the same flow.

## Testing

Against a **cleared install**, which is the state CI actually has:

- before: `Assertion is false: id: searchstop.query is visible`
- after: runs start to finish. The sheet is detected and dismissed on the origin picker, and correctly `SKIPPED` on the destination.

Also confirmed the intermediate state, which is what proves bug 1 was real: with the flow in place but before the `testTagsAsResourceId` fix, the dismiss step still `SKIPPED` because the tag was invisible, and the flow still failed.

`./gradlew detekt --continue`, `testAndroidHostTest --continue` and `fullQualityChecks.sh` all green.

### DIFF
--- a/.maestro/shared/dismiss-map-options.yaml
+++ b/.maestro/shared/dismiss-map-options.yaml
@@ -1,0 +1,44 @@
+# Gets past the first-run map options sheet, and is safe to call on a device that
+# has already seen it.
+#
+# SearchStopViewModel opens this sheet by itself the first time a rider ever
+# reaches the stop search screen, gated on KEY_HAS_SEEN_MAP_OPTIONS_SHEET. While
+# it is up it covers `searchstop.query`, so on a fresh device the search field is
+# not slow to appear, it is unreachable.
+#
+# That is what failed the nightly's 02-plan-trip on every run and survived the
+# timeouts being doubled: waiting longer for a field behind a modal sheet never
+# helps. It passes on a developer emulator because that device saw the sheet
+# once, months ago, and CI is fresh every time. The screenshot Maestro captured
+# at the failure is a full-screen map with this sheet over it.
+#
+# Same shape as dismiss-intro.yaml, and for the same reasons: unconditional at
+# the call site, `optional` here, so a device that has already dismissed it pays
+# one timeout and moves on.
+#
+# Tapping Done rather than the scrim or the drag handle, because Done is what
+# writes KEY_HAS_SEEN_MAP_OPTIONS_SHEET. Dismissing any other way leaves the flag
+# unset and the sheet returns on the next visit, which within one run means the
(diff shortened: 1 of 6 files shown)


ksharma-xyz:
* **#1949** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1949?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1949?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1951, opened 2026-08-22
fix(maestro): get the nightly's iOS leg green

ksharma-xyz:
## What

Both flows failing on the nightly's iOS leg. Reproduced on an iPhone 17 simulator; **neither is an app bug**.

## 06-permissions-denied

Died in under a second, before a single assertion was read. `deny` is Android-only vocabulary for this permission:

```
IllegalArgumentException: wrong argument value 'deny' was provided for 'location' permission
  at util.LocalSimulatorUtils.setLocationPermission(LocalSimulatorUtils.kt:578)
```

It threw out of `launchApp` itself, which is why the flow's actual subject, that the app works with location denied, was never exercised on iOS.

`never` is accepted by **both** drivers. Verified against a simulator and an emulator separately before changing the flow, so this stays one value rather than a platform split.

## 03-rotation-sweep

Excluded from the iOS leg by tag. `setOrientation` does not work on the simulator:

- it reports `COMPLETED`
- the app stays in portrait: the hierarchy bounds after the rotate are still `[0,0][402,874]`, and the screenshot is a fully rendered **portrait** home screen with the search row plainly on it
- the accessibility tree comes back **empty** for that moment (zero resource-ids), so every id assertion after the rotate fails against a screen that is drawn correctly

The app's `Info.plist` lists `UIInterfaceOrientationLandscapeLeft` and `LandscapeRight`, so landscape is supported and this is the driver, not the app.

It would be Android-only regardless. The flow exists to catch **Activity recreation** faults, a route missing from the NavKey `SerializersModule` or state in `remember` rather than `rememberSaveable`, and iOS does not destroy and rebuild the world on rotation. There is no equivalent failure there for it to find.

`--exclude-tags=android-only` on the iOS leg; the Android leg is untouched and still runs it.

## Why tagging rather than deleting

Excluding a flow because it fails is how a suite ends up proving nothing, so the reasoning is written into the flow's own header, with the evidence, rather than left as a bare tag.

## Testing

Local, against the merged `main`:

| | before | after |
|---|---|---|
| iOS `.maestro/nightly/` | 2 of 4 failing | **3 of 3 passed** (rotation excluded) |
| Android `.maestro/nightly/` | — | **4 of 4 passed**, rotation sweep included |

`detekt` green; `actionlint` reports the same 8 pre-existing `SC2086` notes as before, none new.

### DIFF
--- a/.github/workflows/maestro-nightly.yml
+++ b/.github/workflows/maestro-nightly.yml
@@ -221,12 +221,17 @@ jobs:
       # rather than gates, so one retry buys signal without hiding a real break:
       # a genuine regression fails twice. The PR lane gets no retry on purpose —
       # a flake that blocks a merge should be fixed, not papered over.
+      # --exclude-tags=android-only skips 03-rotation-sweep here. setOrientation
+      # is a no-op on the simulator: it reports success, the app stays portrait,
+      # and the accessibility tree empties for that moment, so the flow fails
+      # against a screen that is drawn correctly. See that file's own header.
       - name: Run smoke and nightly flows
         run: |
           set -o pipefail
           run_suite() {
             maestro --device "${{ steps.sim.outputs.udid }}" \
-              test -e APP_ID=xyz.ksharma.krail .maestro/smoke/ .maestro/nightly/
+              test --exclude-tags=android-only \
+              -e APP_ID=xyz.ksharma.krail .maestro/smoke/ .maestro/nightly/
           }
           if ! run_suite; then
             echo "::warning::Maestro suite failed on iOS; retrying once."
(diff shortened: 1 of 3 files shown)


ksharma-xyz:
* **#1951** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1951?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1951?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1952, opened 2026-08-22
docs(learning): record the nightly lane and the platform-drift bugs

ksharma-xyz:
## What

Nothing from today's work was in `docs/learning/`, and most of it qualifies under that log's own rules: detekt and tests were green while the bugs were live, the first fix was wrong and so was the second, and diagnosis needed artifacts rather than reading code.

## Entries

**[A lane that never ran](docs/learning/2026-08-22-a-lane-that-never-ran.md)** — the Maestro nightly. Three stacked bugs, each hiding the next: the lane never discovered a flow and exited 1 anyway (so it looked like a failing suite), no `testTag` inside any `ModalBottomSheet` was ever published as a resource id, and a first-run sheet covered the field a flow was waiting for.

**[Two platforms, two vocabularies](docs/learning/2026-08-22-two-platforms-two-vocabularies.md)** — the four Ask KRAIL bugs, reported as unrelated, all one shape: a contract both platforms are meant to honour where only one does, and nothing fails when the other does not.

Both entries lead with the **wrong turns**, per the log's rules. Mine, in this case:

- assuming the two Maestro lanes failed the same way, which reached a merged PR description before anyone checked
- doubling flow timeouts, getting an identical failure, and reading that as a reason to increase them again rather than as proof the wait was never the problem
- naming "the NSW API" as a cause for an Android failure whose screenshot was solid black with an empty hierarchy

## The check

`AiUnavailableReasonCoverageTest` walks `AiUnavailableReasons` and fails if any value falls through to the generic banner, or decides the wrong thing about hiding the entry point. It also pins the vocabulary size, so adding a constant without deciding what it says to a rider fails.

**Mutation-checked before committing**: reverting the `DEVICE_UNSUPPORTED` arm to `null` fails it with the right message. The pre-existing test it supplements asserted the *phase* and stayed green for months while riders read the wrong sentence, which is exactly the failure mode the entry describes.

## A note on the guard that reviewed this PR

`LearningLogActionsTest` rejected my first draft: a ticked box named `` `dismiss-intro.yaml` `` and that does not resolve from the repo root. Fixed to the full path rather than loosening the guard. That test is the reason a ticked box means anything, and it did its job on the person adding entries.

Also adds the 2026-08-21 entry to the index, which was written without being listed.

## Testing

`./gradlew detekt --continue` and `./gradlew testAndroidHostTest --continue` both green, including `LearningLogActionsTest`.

### DIFF
--- a/docs/learning/2026-08-22-a-lane-that-never-ran.md
+++ b/docs/learning/2026-08-22-a-lane-that-never-ran.md
@@ -1,0 +1,104 @@
+# The Maestro nightly had never run a single flow, and looked like a failing test suite
+
+**2026-08-22** · **CI / end-to-end lane** · **Cost: four dispatched nightly runs, two wrong diagnoses, one unnecessary PR**
+
+## Symptom
+
+The nightly lane was red every run. The PR smoke lane was red on every pull request,
+regardless of what the pull request changed. Both had been red for days and people had
+learned to scroll past them.
+
+The visible failure on the smoke lane was:
+
+```
+[Failed] 02-plan-trip (44s) (Assertion is false: id: searchstop.query is visible)
+```
+
+The nightly's job log ended without ever naming a flow.
+
+## Root cause
+
+Three separate bugs, stacked. Each one hid the next.
+
+1. **The nightly never started.** It passes `.maestro/` so it can cover the whole tree, and
+   Maestro does not recurse into subdirectories. Pointed at a directory whose flows all live
+   one level down it finds nothing and exits **non-zero**:
+
+   ```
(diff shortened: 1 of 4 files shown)


ksharma-xyz:
* **#1952** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1952?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1952?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1954, opened 2026-08-23
fix(searchstop): open on the list, restore the reverse-stops button

ksharma-xyz:
## What

Two fixes on the stop-search path, plus the refactor one of them needed and a triage
doc for the Maestro lane.

Tapping From or To opened stop search on the map instead of the list, on both
platforms. Separately, the action column beside From and To left its upper slot empty
on a device that cannot run on-device AI, so Search moved up into it.

## Changes

**Stop search opens on the list** (`SearchStopScreen.kt`)
- `showMap` back to `false`. It was flipped to `true` in `8122380db`, a commit about
  stop labels with no reason to touch it.
- The `LaunchedEffect` beside the flag hides the keyboard and frees field focus while
  the map is up, so map-first meant landing on a screen that could not be typed into
  until the field was tapped. `MapAutoInitEffect` also built MapLibre on every visit.
- Only the single-pane layout reads the flag. `SearchStopScreenDualPane` always draws
  list left and map right, which is why tablet and phone landscape were correct.
- The first-run map-options sheet is unaffected: `showMapOptionsOnOpen` is consumed by
  `MapOptionsFirstTimeShown`, which the map fires, so it now appears the first time a
  rider actually opens the map.

**Reverse-stops button** (`SearchStopRow.kt`)
- The upper action slot always holds a button now: Ask KRAIL where `isWayInAvailable`
  is true, reverse stops where it is not. Search stays in the lower slot either way.
- Restores what `696ccd1d9` removed when the wheel took that slot: `ReverseStopClick`,
  the ViewModel branch, `StopResultsManager.reverseSelectedStops` with its
  implementation and fake, and the `reverse_stop_click` event.
- The icon half-turns per tap and stays there, so the arrow points the way the stops
  now run. The resting angle is `rememberSaveable`, so rotation does not reset it.
- No `ANALYTICS_REGISTRY_HANDOFF.md` row: `reverse_stop_click` is already listed as
  Registered, same name, no params.

**ViewModel size** (`SavedTripsInfoTiles.kt`, new)
- `SavedTripsViewModel` was at detekt's `LargeClass` limit, so the reverse branch could
  not be added. Info-tile handling moves out as extension functions; the members they
  use widen from `private` to `internal`.
- `onParkRideCardClick` split into expanded/collapsed halves to clear `LongMethod`. The
  two are kept separate rather than merged with the set operation flipped: collapsing
  has no "spots have arrived, stop showing loading" arm.
- No behaviour change.

**Tests**
- `singlePane_opensOnTheList_evenWhenAMapIsAlreadyInitialised` uses `MapUiState.Ready()`.
  This is the gap that let the regression ship: the list renders whenever `mapUiState`
  is null, so every existing fixture passed identically with the flag either way.
  Mutation-checked, it fails with the flag flipped back.
- `selectOnMap_isOfferedInTheList_whenMapsAreAvailable` and
  `selectOnMap_asksForTheMap_whenTapped` hold the way into the map, so a fix that just
  pinned the list would not pass.
- Two `SavedTripsViewModelTest` cases for reversing: both stops set, and the half-filled
  pair where the null has to travel with the swap.
- `02-plan-trip` and `03-rotation-sweep` assert `searchstop.recents` right after tapping
  the field, on both routes into the screen.

**Maestro flows**
- The two `dismiss-map-options.yaml` calls in `02-plan-trip` are removed. That sheet is
  opened by the map, which is now opt-in, so they were two dead ten-second waits in the
  one flow whose history is timing.
- `03-rotation-sweep` had no such call and passed only because `02-plan-trip` ran earlier
  in the same invocation and dismissed the sheet. That order dependency is gone.
- The shared flow stays, with a header saying when it is needed again.

**Docs**
- `docs/MAESTRO_TRIAGE.md`: artifact layout, the order to read it in, what the run does
  not capture (no video, no frames on a passing step, no device log on iOS), and a table
  of failure signatures. Linked from `CLAUDE.md`, which had no mention of Maestro.
- `SEARCH_STOP_UX.md` gains the landing-surface rule it had only implied.
- `ASK_KRAIL_UX.md` gains the never-empty-slot rule.
- Learning entry for the landing-surface bug.

## Testing

- `./scripts/fullQualityChecks.sh` green (layout invariants, Android compile, iOS
  Simulator compile, detekt).
- `./gradlew :feature:trip-planner:ui:testAndroidHostTest :core:analytics:testAndroidHostTest
  :composeApp:testAndroidHostTest --continue` green.
- Maestro on a `pm clear`'d emulator, which is CI's starting state:
  `02-plan-trip` and `03-rotation-sweep` both green, every step completed.
- Device QA on emulator, Pixel 10 Pro and iPhone 11: install, open, reverse a pair of
  stops, rotate portrait/landscape with data loaded, dark/light toggle, no
  `FATAL EXCEPTION` in logcat. Node bounds confirm Search stays in the lower slot on
  both the Ask KRAIL and reverse branches.

## Snapshots

| Screen | Before | After |
|---|---|---|
| SearchStopScreen, tapping From | opened on the map, field unfocused | opens on the list, field focused, keyboard up |
| SearchStopRow, no on-device AI | upper slot empty, Search beside From | reverse button in upper slot, Search beside Destination |

Images not attached: they cannot be uploaded to a PR body from the CLI. Local captures
are in the session scratchpad if you want to drag them in.


ksharma-xyz:
* **#1954** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1954?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1954?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1955, opened 2026-08-23
fix(speech): end iOS sessions on silence and drop blank transcripts

ksharma-xyz:
## What

An iOS listening session in Ask KRAIL would not end when the rider stopped talking, and when it did end it cleared the sentence the rider had just watched arrive. Both were in code that only exists on iOS, because `SFSpeechAudioBufferRecognitionRequest` has no notion of a speaker finishing: `isFinal` arrives only after the app calls `endAudio()`. Android's `SpeechRecognizer` owns that decision itself.

## Changes

**Blank transcripts (all platforms)**

- `SpeechToTextResult` documents the contract: `Partial` and `Final` always carry words. A session that heard nothing ends as `Error(NO_RESULT)` rather than as a result that is nothing.
- `AndroidSpeechToTextService` treats a blank recognition candidate as no candidate. It could not previously produce one, but only because the candidate is read through a nullable helper.
- `AiSearchInputViewModel` refuses to write a blank transcript into the field whichever platform produced it, for partials and finals alike. The field is what the rider is looking at, so a blank write is their sentence being erased.

**Session ending (iOS)**

- `stopFeedingAudio()` is now the one way to finish listening: microphone off, tap removed, then `endAudio()`. The silence watcher previously called `endAudio()` alone while the `AVAudioEngine` tap kept appending buffers to the same request, so the recogniser had no reason to finish and the ViewModel's ten second backstop became the only thing ending a session. The rider's stop button always did this correctly; the watcher, which ends nearly every session, did not.
- The quiet window restarts only when the transcript gains a word. It previously restarted on any changed transcript, and the recogniser goes on revising for seconds after the rider stops (capitalisation, homophones, punctuation), so a finished sentence held the microphone open on the strength of the recogniser tidying up after itself.
- `SILENCE_THAT_ENDS_THE_SESSION_MILLIS` is 3000, was 4000. Android sets two windows and the one that fires for a trip sentence is the 3000ms complete-sounding-phrase window, so matching the other one cost about a second on every sentence. The constant carries a note that the two platforms measure different events (Android measures silence, iOS measures the last word arriving) and that equal numbers are a coincidence, not a contract.
- The session rules moved into `TranscriptWatch` in `commonMain`. `iosMain` has no host test task in this project, which is why none of this had coverage.

**Left untouched**

- The ViewModel's 10s and 15s ceilings stay. They are the correct backstop for a recogniser that never reports an end; the fault was that they had become the normal path.
- The `SFSpeechRecognizer` path itself is unchanged. `SpeechAnalyzer` with `DictationTranscriber` and `SpeechDetector` is the better fit for this surface and is a separate piece of work: it needs a Swift shim, since the API is Swift only.

## Testing

- `./gradlew :core:speech-to-text:testAndroidHostTest :feature:trip-planner:ui:testAndroidHostTest --continue` green. `TranscriptWatchTest` is new, 10 tests. `AiSearchInputViewModelTest` is 54, three of them new.
- `LongMethod` on `startListening` was resolved by extracting `TranscriptWatch`, not suppressed.
- Device QA on iOS was not possible here. The Ask KRAIL entry point is gated on on-device AI capability, so the surface does not open on the iOS simulator or the Android emulator. Both apps build, install and launch, and the emulator log has no `FATAL EXCEPTION`, but the microphone path itself has not been exercised on hardware.

### DIFF
--- a/core/speech-to-text/src/androidHostTest/kotlin/xyz/ksharma/krail/core/speechtotext/TranscriptWatchTest.kt
+++ b/core/speech-to-text/src/androidHostTest/kotlin/xyz/ksharma/krail/core/speechtotext/TranscriptWatchTest.kt
@@ -1,0 +1,142 @@
+package xyz.ksharma.krail.core.speechtotext
+
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertNull
+import kotlin.test.assertTrue
+import kotlin.time.Duration.Companion.seconds
+import kotlin.time.TestTimeSource
+
+/**
+ * The two rules the iOS speech surface turns on, tested here because iOS source has no test task
+ * in this project. Both were rider-facing bugs: a sentence cleared from the field, and a session
+ * that never ended on its own.
+ */
+class TranscriptWatchTest {
+
+    @Test
+    fun `partials carry the words as they arrive`() {
+        val watch = TranscriptWatch()
+
+        assertEquals(SpeechToTextResult.Partial("central"), watch.record("central", isFinal = false))
+        assertEquals(
+            SpeechToTextResult.Partial("central to town hall"),
+            watch.record("central to town hall", isFinal = false),
+        )
+        assertTrue(watch.heardAnything)
+    }
+
+    @Test
+    fun `a blank partial is not sent on`() {
+        val watch = TranscriptWatch()
+        watch.record("central to town hall", isFinal = false)
+
+        // Nothing to send, rather than an empty string to write into the rider's field.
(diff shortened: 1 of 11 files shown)


ksharma-xyz:
* **#1955** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1955?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1955?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1957, opened 2026-08-23
feat(speech): listen through SpeechAnalyzer on iOS 26, SFSpeechRecognizer below

ksharma-xyz:
## What

Replaces the iOS speech path with iOS 26's `SpeechAnalyzer`, falling back to `SFSpeechRecognizer` on anything older.

`SFSpeechRecognizer` is a transcription **stream**: it never decides a speaker has finished, because `isFinal` arrives only after the app calls `endAudio()`. Every end-of-speech decision therefore had to be inferred from how text happened to arrive, and every fault in [the 2026-08-23 learning entry](../blob/main/docs/learning/2026-08-23-the-blank-transcript-that-cleared-the-field.md) came from that inference. iOS 26 ships the two missing pieces as modules: `DictationTranscriber` (short queries, same on-device model) and `SpeechDetector` (voice activity detection, reported in audio time).

Two commits, kept separate: the bridge and its rule, then the wiring.

## Changes

**`feat(speech): add an iOS SpeechAnalyzer bridge and its end-of-speech rule`**

- `spmForKmp` with `cinteropName = "speechBridge"`, matching how `:core:ai-text` declares its bridge. `SpeechAnalyzer` is a Swift `actor` with `AsyncSequence` results and no Objective-C surface, so cinterop cannot see it.
- `src/swift/speechBridge/SpeechBridge.swift` exposes `checkAvailability`, `start`, `finish` and `cancel` over plain closures and primitives. It reports what Apple's modules say, including each voice-activity answer with the audio time it applies to, and decides nothing.
- `SpeechActivityWatch` in `commonMain` owns the decision instead: how long the rider has been quiet, and whether that is enough. 3000ms once speech has been heard, 6000ms before any.
- `SpeechActivityWatchTest`, 7 cases.

**`feat(speech): listen through SpeechAnalyzer on iOS 26, SFSpeechRecognizer below`**

- `SpeechAnalyzerSpeechToTextService` drives the analyzer. There is **no silence-polling coroutine any more**: the detector reports, and each report is enough to decide on.
- It reuses `TranscriptWatch`, so the rule that a blank transcript is never sent on as one is written once and both iOS paths get it.
- `PreferredSpeechToTextService` picks between them inside `checkAvailability`, the call every caller already makes before listening. In `commonMain`, because choosing between two implementations is not an iOS idea and `iosMain` has no test task.
- When the analyzer cannot run, the fallback's availability answer is the one reported. The two do not share a set of reasons, and reporting "the locale's model is still downloading" from a device whose older recogniser will listen right now would take the feature away for no reason.
- `PreferredSpeechToTextServiceTest`, 5 cases.
- README rewritten: it claimed neither platform actual was implemented, and that iOS needed no SPM bridge. Neither is true any more.

**Why the split between Swift and Kotlin is where it is.** The bridge owns mechanism, `commonMain` owns policy. Swift source in this project has no test task, which is how three rider-facing faults lived in the `SFSpeechRecognizer` path unnoticed.

**On the window being 3000ms again:** same number as Android's complete-sounding-phrase window, and for the first time the same measurement. It previously counted seconds without a transcript update, which is recognition lag plus silence; now a real detector reports real silence.

**Permissions change for riders on iOS 26:** `SpeechAnalyzer` needs the microphone alone, where `SFSpeechRecognizer` also needs speech-recognition authorization, so that path shows one system prompt instead of two. Both Info.plist keys stay, because the legacy path is still live below iOS 26.

**Left untouched:** `IosSpeechToTextService` keeps its name. Renaming it to say "legacy" would have rippled into a merged learning-log entry that records what was true when it was written.

## Testing

- `./gradlew testAndroidHostTest --continue` green across all modules. 26 tests in `:core:speech-to-text`, 12 of them new.
- `xcodebuild` of `iosApp` for the simulator succeeds with the second SPM package present; the app installs, launches and renders the home screen with no crash.
- The Swift is type-checked by the cinterop build against the iOS 26.5 SDK (Xcode 26.6).

**Not verified on hardware, deliberately merged anyway.** No microphone has been through the analyzer: the Ask KRAIL surface is gated on on-device AI, so it does not open on a simulator, and there is no iOS device on hand. The audio-format conversion, the detector's reporting cadence, and whether `finalizeAndFinishThroughEndOfInput()` yields a transcript in practice are reasoned from Apple's documentation and held up by the compiler, not by a device. Verification happens through TestFlight after this lands, which is why it lands as two readable commits rather than waiting. The fallback contains the blast radius: anything short of `Available` from the analyzer routes back to the `SFSpeechRecognizer` path.

### DIFF
--- a/core/speech-to-text/README.md
+++ b/core/speech-to-text/README.md
@@ -5,9 +5,9 @@ On-device speech-to-text for the "Speak" tab of the AI search-input flow
 Wraps each platform's own speech recognizer behind [`SpeechToTextService`](src/commonMain/kotlin/xyz/ksharma/krail/core/speechtotext/SpeechToTextService.kt),
 the same `expect`/`actual`-per-platform shape as `core:ai-text`'s `AiTextService`.
 
-**Status: interface + module boundary only. Neither platform actual is implemented yet** —
-both `AndroidSpeechToTextService` and `IosSpeechToTextService` always report `Unavailable`.
-See the doc comment on each actual for exactly what real implementation needs.
+**Status: both platforms implemented.** iOS has two, chosen at runtime by
+`PreferredSpeechToTextService`: `SpeechAnalyzerSpeechToTextService` on iOS 26 and
+`IosSpeechToTextService` (`SFSpeechRecognizer`) everywhere else. See "The two iOS paths" below.
 
 ## Why a new module, not an existing library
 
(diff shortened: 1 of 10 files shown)


ksharma-xyz:
* **#1957** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1957?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1957?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1958, opened 2026-08-24
fix(searchai): add spoken words to the field instead of replacing it

ksharma-xyz:
## What

Two pieces of device feedback on Ask KRAIL.

1. Speaking replaced whatever was already in the field. Now it adds to it.
2. The way in was the AI wheel. Now it is a mic wearing the gradient as a ring.

Both are shared code, so Android and iOS behave identically.

## Changes

**`fix(searchai): add spoken words to the field instead of replacing it`**

- The field's contents are captured when listening starts, and every partial and final is joined onto them. Captured per session, so a second session adds to what the first left rather than to a stale copy, and cleared wherever the draft is thrown away.
- `joinSpokenText` is a top-level function in its own file. `AiSearchInputViewModel` is at its `TooManyFunctions` limit and a local function would not have helped; a pure string join also tests directly.
- `SpokenTextJoinTest` covers the empty field, a whitespace-only field, and whitespace already present on either side of the join. Each of those is something the rider sees, as either a leading space or two words run together.
- Four new `AiSearchInputViewModelTest` cases for the behaviour end to end.

*Why adding rather than replacing.* The field is the only copy of what the rider typed, so replacing destroys input with no way back. Every mic that lives inside a text field adds to what is there, from the iOS keyboard's own dictation to Gboard's, because a mic beside a keyboard is another way to type rather than a different way to ask. Replacing has one case going for it, a rider re-tapping the mic to redo a mis-heard sentence, where adding gives them the sentence twice. That is the worse reading of the tap but the better failure: it is on screen, obviously wrong, and one gesture to clear. Losing typed words is silent.

**`feat(taj): open Ask KRAIL from a mic with a turning gradient ring`**

- `AiMicButton` keeps the glyph and the fill in the theme's own content and surface colours and puts the gradient in a ring around the button's edge, so the control reads as a mic first and an AI surface second.
- The ring is drawn at rest and turns for the same one beat the wheel did: the settle after the dialog closes onto the row with an answer.
- It lives in `:taj`, where `ButtonTokens.RoundButtonSize` is visible, so the ring's radius comes from the same token the button sizes itself with and the two cannot drift into an oval.
- Ring stroke is `Regular` rather than the modifier's `Thick` default, which is tuned for a card's edge. At 4dp on a 48dp circle the ring stops being a border and becomes the button.

*Why not the wheel.* The wheel is KRAIL's mark for *on-device AI produced this*, which is right on the alert-summary card and wrong on a door: on the way in it named the technology to a rider looking for a way to say where they are going. `AiWheelMark` is unchanged and still used by the alert-summary card.

## Testing

- `./gradlew testAndroidHostTest --continue -PciQuality=true` green across all modules. `AiSearchInputViewModelTest` 58 cases, `SpokenTextJoinTest` 6 new.
- `./gradlew verifyTestWiring verifyTestingModuleUsage verifyNoAdHocBoundaryFakes verifyIosTestClassification --continue -PciQuality=true` green.
- `TooManyFunctions` on the ViewModel was resolved by extracting to a separate file, not suppressed.
- Screenshots recorded for the new component. No existing screenshot changed.

## Snapshots

The Ask KRAIL entry point is gated on on-device AI, so it does not render on a simulator at all. These previews are the only way to see it.

| Screen | Light | Dark |
|---|---|---|
| AiMicButton, at rest (theme gradient) |  |  |
| AiMicButton, cool gradient |  |  |

### DIFF
--- a/feature/trip-planner/ui/src/commonMain/kotlin/xyz/ksharma/krail/trip/planner/ui/components/SearchStopRow.kt
+++ b/feature/trip-planner/ui/src/commonMain/kotlin/xyz/ksharma/krail/trip/planner/ui/components/SearchStopRow.kt
@@ -58,7 +58,7 @@ import krail.feature.trip_planner.ui.generated.resources.ic_search
 import org.jetbrains.compose.resources.painterResource
 import xyz.ksharma.krail.taj.LocalContentColor
 import xyz.ksharma.krail.taj.LocalThemeColor
-import xyz.ksharma.krail.taj.components.AiWheelMark
+import xyz.ksharma.krail.taj.components.AiMicButton
 import xyz.ksharma.krail.taj.components.Button
 import xyz.ksharma.krail.taj.components.ButtonDefaults
 import xyz.ksharma.krail.taj.components.RoundIconButton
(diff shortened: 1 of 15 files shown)


ksharma-xyz:
* **#1958** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1958?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1958?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1959, opened 2026-08-24
fix(searchai): fall back to Home for the From stop, and document both ladders

ksharma-xyz:
## What

Reported on iOS: "I want to go to `<somewhere>`" filled the destination and left **From** blank, even with location available.

Two commits. The fix, and the docs catch-up that should have shipped with the mic change.

## The bug

The fallback that fills From needs a location, and **every way of not having one arrives as the same `null`**: permission denied, restricted, or the fix timing out. All three produced the same blank field with nothing to tell them apart, which is why the report could not be narrowed down.

Rather than guess which, this does two things: adds a last resort so the common case stops being blank at all, and logs each branch so the next build says definitively which one happened.

## How the From stop is decided now

Ordered by how much each signal knows about **right now**. A guess never overrules something known.

| | Signal | Knows |
|---|---|---|
| 1 | What the rider said | Their intent, certainly |
| 2 | A labelled stop within walking distance | Their position. The label is the platform they actually use, where the nearest stop is often a shelter nobody uses |
| 3 | The nearest stop | Their position |
| 4 | **Home, if set** | Only their habit |
| 5 | Blank | Nothing |

**Location always beats Home**, and that ordering is the whole design. A rider at work asking for the airport is served correctly by where they are and wrongly by where they usually start. Home is reached only when there is no location at all.

**Two cases deliberately stay blank rather than reaching for Home**: standing at the destination, and a known location with no stop near it. Both are things the app knows about the rider right now, and a guess must not contradict them.

Filling From in is answering a question the rider asked, not inventing one: it happens only when a destination was actually understood. A sentence that resolved to no places still takes no fallback, which is the existing failure mode 5.

## Changes

- `LabelledStopLocator.homeStop` reads the Home label. Home is the only label this app treats as permanent (it cannot be renamed, only reassigned or cleared), which is what makes keying behaviour on it safe.
- `HOME_STOP_LABEL` replaces a second spelling of `"home"` that had grown in `AiGreeting`, so the two readers cannot drift.
- Every branch logs under `[AI_ORIGIN]`, including **why** a location was unavailable, unwrapped at the one place the `Result` still exists.
- `originStop` restructured to one return per function rather than suppressing `ReturnCount`.
- Docs: `AI_SEARCH_UX.md` gains this ladder and the end-of-speech ladder; `ASK_KRAIL_UX.md` gains the append-on-speak rule and says mic rather than wheel; `ASK_KRAIL_CAPABILITIES.md` no longer describes a wheel riders cannot see.

## Testing

- `RiderOriginLocatorTest`, 7 new cases covering the full ladder: Home used when location is missing, location beating Home when both exist, Home refused when it is also the destination, and both deliberate blanks.
- `./gradlew testAndroidHostTest --continue -PciQuality=true` green across all modules.
- `./gradlew verifyTestWiring verifyTestingModuleUsage verifyNoAdHocBoundaryFakes verifyIosTestClassification --continue -PciQuality=true` green.
- `./scripts/fullQualityChecks.sh` green.

**Not verified on a device.** The reported symptom is iOS and cannot be reproduced here (Ask KRAIL does not open on a simulator). The logging is what will confirm the original cause; the fallback makes the common case correct regardless of which it was.

### DIFF
--- a/docs/marketing/ASK_KRAIL_CAPABILITIES.md
+++ b/docs/marketing/ASK_KRAIL_CAPABILITIES.md
@@ -75,8 +75,8 @@ the other is left to them.
 ## What it looks like
 
 The surface is painted from the rider's chosen theme: two colours from that theme drifting
-behind the input, the same pair on the AI wheel, the border, and the search screen's background.
-Change the theme and all of it follows.
+behind the input, the same pair on the mic ring that opens it, the border, and the search
+screen's background. Change the theme and all of it follows.
 
 ---
 
--- a/feature/trip-planner/ui/AI_SEARCH_UX.md
+++ b/feature/trip-planner/ui/AI_SEARCH_UX.md
@@ -30,7 +30,7 @@ Every way this can fail, what the rider gets, and whether a test holds it in pla
 
 | # | Failure | Rider sees | Test |
 |---|---|---|---|
-| 1 | Feature flag off | **No way in.** The wheel is not drawn | `with the flag off there is no way in`, `with the flag on the way in is there and opens`, `starting over keeps the way in`, plus `submit while flag is off does nothing` |
+| 1 | Feature flag off | **No way in.** The mic is not drawn | `with the flag off there is no way in`, `with the flag on the way in is there and opens`, `starting over keeps the way in`, plus `submit while flag is off does nothing` |
 | 2 | Model not downloaded, or downloading | "still downloading… try again in a moment" | `model downloadable lands in DOWNLOADING`, `model downloading lands in DOWNLOADING too` |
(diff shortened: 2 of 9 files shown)


ksharma-xyz:
* **#1959** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1959?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1959?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Issue #1963, opened 2026-08-28
Fuzzy stop search prefilter drops the query's most distinctive word


ksharma-xyz:
## Summary

The fuzzy stop search prefilter spends its entire fragment budget on the first word of a
multi-word query, so the most distinctive word is never queried against the database. Stops
that the ranker would score perfectly never enter the candidate pool, and the screen shows
"No match found".

This is live. `enable_fuzzy_stop_search` is served on.

## Background: two stages, and only one of them can be recovered from

Scoring all NSW stops per keystroke is far too slow, so `fetchFuzzyCandidates`
(`RealStopResultsManager.kt`) narrows to `MAX_FUZZY_CANDIDATES = 200` first, and only those
are ranked. **A stop the prefilter misses can never be shown, however good a match it is.**

## The defect

`RealStopResultsManager.kt:173-189` builds up to four sliding trigrams per token, flattens
them token by token, then truncates:

```kotlin
}.distinct().take(MAX_PREFIX_QUERIES)   // = 4
```

Because `flatMap` emits token-major, the truncation slices inside the first token. For
`"wollongong central burell"`:

```
built:   [wol, oll, llo, lon,  cen, ent, ntr, tra,  bur, ure, rel, ell]
kept:    [wol, oll, llo, lon]            <- all from "wollongong"
dropped: everything for "central" and "burell"
```

`%burell%` matches a handful of stops statewide. `%wol%` matches hundreds. The single most
selective fragment in the query is the one guaranteed to be discarded.

## Why the pool then fills with the wrong stops

Each surviving fragment contributes up to `MAX_CANDIDATES_PER_PREFIX = 50`, and
`selectProductClassesForStop` has no `ORDER BY`, so rows arrive in feed order and an
arbitrary 50 win the cap. "Wollongong Central, Burelli St" lost its slot to entries such as
"Murphys Ave opp Wollongong Botanic Garden".

## Reproduction

Only via the fuzzy path, since the token-wise LIKE change fixed the exact path for this
particular query. A query where no whole token matches any stop name still reproduces it:

- `burell st wollongong` (tokens reordered, so the order-preserving exact path cannot match)

## What was measured

A recall harness over all NSW stops in `nsw_stops_eval.csv`, checking whether the intended
stop reaches the candidate pool at all, separately from how it is then ranked. 28 queries:

| Prefilter strategy | Recall |
|---|---:|
| Token-major x4 (shipped) | 27/28 |
[...]

PR #1964, opened 2026-08-28
fix(search): match stop names token-wise so the comma need not be typed

ksharma-xyz:
## What

Stop name matching ignored the punctuation in NSW GTFS names, so a rider had to type the
comma to find any multi-word stop. `"wollongong central burell"` returned nothing while
`"wollongong central, burell"` returned the stop and all its stands.

## Changes

- `stopNameLikePattern` (new, `:sandook`) splits a query on whitespace and punctuation and
  rejoins the tokens with `%`, so the wildcard spans whatever separators the feed uses.
  `"wollongong central burell"` becomes `%wollongong%central%burell%`.
- `selectProductClassesForStop` binds that pattern through an `ESCAPE` clause instead of
  concatenating `'%' || :stopName || '%'`. This also closes a hole where a typed `%` or `_`
  acted as a SQL wildcard and matched every stop in the state.
- Both sandook implementations build the pattern, so every caller is covered. The `stopId`
  operand stays raw, since it is an exact-equality match rather than a pattern.
- Matching stays order-preserving, so this does not widen results the way an unordered
  AND-of-LIKEs would. Reordered tokens remain the fuzzy ranker's job.
- `RemoteConfigDefaults` had `enable_fuzzy_stop_search` defaulting to `false` while Remote
  Config serves `true`, which made the shipped behaviour unreadable from source. Corrected,
  with a comment on why the two must agree. Defaults only apply before the first fetch, so
  this changes nothing for an install that reaches Remote Config.

Left untouched: the fuzzy trigram prefilter drops a multi-word query's most distinctive word
before it reaches the database. Investigated during this change and filed separately as
#1963, because the prefilter and the ranker's score threshold are coupled and every strategy
that improved candidate recall regressed ranking on another tracked query.

## Testing

- `NswStopsNameSearchTest` (new, 8 cases) runs against real in-memory SQLite rather than a
  fake, because only SQLite's own `LIKE` shows the defect. Confirmed as a genuine regression
  test: 4 of the 8 fail against the previous predicate.
- `StopSearchPipelineReproTest` drives the real pipeline over the shipped NSW stop set. Its
  fake DB still hardcoded contiguous substring matching, which no longer describes the SQL,
  so it now uses ordered-token containment. Two scenarios added:
  `"wollongong central burell"` and `"central station eddy"`, both resolving in the top five.

Device QA on `emulator-5554`, debug variant:

| Query | Before | After |
|---|---|---|
| `wollongong central burell` | "No match found!" | Burelli St plus Stands A/B/D/E |
| `wollongong central, burell` | worked | identical results |
| `burwood`, `central` | worked | unchanged |

Rotated with results loaded: query text and results survived, no `FATAL EXCEPTION` in logcat.

### DIFF
--- a/core/remote-config/src/commonMain/kotlin/xyz/ksharma/krail/core/remoteconfig/RemoteConfigDefaults.kt
+++ b/core/remote-config/src/commonMain/kotlin/xyz/ksharma/krail/core/remoteconfig/RemoteConfigDefaults.kt
@@ -161,9 +161,14 @@ object RemoteConfigDefaults {
                 first = FlagKeys.TRIP_TRACKING_ENABLED.key,
                 second = true,
             ),
+            // Matches the value served by Remote Config. These defaults only apply before the
+            // first fetch completes, so a default that disagrees with the served value makes
+            // the shipped behaviour unreadable from source: fuzzy search reads as off here
+            // while every install actually runs it. Keep this in step when the served value
+            // changes, and read `fetchFuzzyCandidates` as live code, not a dormant path.
             Pair(
                 first = FlagKeys.ENABLE_FUZZY_STOP_SEARCH.key,
-                second = false,
+                second = true,
             ),
             Pair(
                 first = FlagKeys.ENABLE_PROTO_BFF.key,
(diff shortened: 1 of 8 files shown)


ksharma-xyz:
* **#1964** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1964?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1964?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1965, opened 2026-08-28
fix(alerts): key service alerts on a stable id instead of the list index

ksharma-xyz:
## What

Yes, this is a real crash, and it is live in `v1.26.0`.

NSW sends distinct alerts under a single heading. The Service Alerts sheet keyed its
`LazyColumn` on the heading alone, so the second alert with a repeated heading threw:

```
java.lang.IllegalArgumentException: Key "trackwork" was already used. If you are using
LazyColumn/Row please make sure you provide a unique key for each item.
```

`652e125bf` prefixed the list index in, which stops the crash. That commit is on `main` but
is in no tag and not on `prod/1.26.0`, so the shipped app still carries the crashing
expression:

```kotlin
key = { _, item -> item.heading.lowercase() }   // v1.26.0
```

The index prefix is unique, but it makes identity positional: reordering or inserting an
alert renumbers every key after it, so Compose treats untouched rows as new ones and drops
their animation and state. This replaces it with a stable id.

## Changes

- `ServiceAlert.alertId` is the `(heading, message)` pair. Alerts reach the UI as a `Set` of
  this data class, so the pair is distinct by construction and, unlike the index, survives
  reordering. Computed rather than stored, so it stays out of the serialized form.
- The `LazyColumn` key is now `"alert_${item.alertId}"`, prefixed to keep it clear of the
  sheet's static item keys (`title_bar`, `ai_alert_summary`, `bottom_spacing`).
- The expanded-row state uses the same id. It previously used `hashCode()`, an `Int` over
  unbounded text, so two alerts could collide and expand together. `expandedAlertId` changes
  from `Int?` to `String?`.

## Testing

`ServiceAlertScreenIdentityTest` (new, 6 cases) renders the real screen with three alerts
that share the heading "Trackwork":

- all three render rather than crashing
- expanding one does not expand its same-heading siblings
- the expanded alert survives `emulateSavedInstanceStateRestore`, which covers the
  `rememberSaveable` type change
- a single alert still renders, and `alertId` separates same-heading alerts while staying
  equal across rebuilt instances

Confirmed these catch the defect rather than merely passing: reverting the key to the
shipped `item.heading.lowercase()` fails 2 of the 6 with the exact
`Key "trackwork" was already used` exception.

Device QA is incomplete and worth calling out: the app was installed and driven to a live
Central to Parramatta timetable with no crash, but no route carried a service alert at the
time, so the changed sheet could not be opened on device. The recreation case that rotation
would have exercised is covered by `StateRestorationTester` above instead.

## Follow-up, not in this PR

The fix is absent from `prod/1.26.0`, so riders on the current release still hit the crash.
Cherry-picking either this change or `652e125bf` onto that branch is a separate call.

### DIFF
--- a/feature/trip-planner/state/src/commonMain/kotlin/xyz/ksharma/krail/trip/planner/ui/state/alerts/ServiceAlert.kt
+++ b/feature/trip-planner/state/src/commonMain/kotlin/xyz/ksharma/krail/trip/planner/ui/state/alerts/ServiceAlert.kt
@@ -9,6 +9,18 @@ data class ServiceAlert(
     val message: String,
 ) {
 
+    /**
+     * Stable identity for one alert, unique within any collection of alerts.
+     *
+     * The feed gives an alert no id, and the heading alone is not unique: NSW sends distinct
+     * alerts under one heading, which is how a `LazyColumn` keyed on heading crashed with
+     * "Key was already used". Since alerts reach the UI as a `Set` of this data class, the
+     * (heading, message) pair is distinct by construction, so it is the identity.
+     *
+     * Computed rather than stored, so it stays out of the serialized form.
+     */
+    val alertId: String get() = "$heading|$message"
+
     fun toJsonString() = Json.encodeToString(serializer(), this)
 
     @Suppress("ConstPropertyName")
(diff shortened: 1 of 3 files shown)


ksharma-xyz:
* **#1965** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1965?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1965?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1966, opened 2026-08-28
fix(search): prefix search result keys and guard product class uniqueness

ksharma-xyz:
## What

Two collision risks found while auditing every Lazy key after the service alert crash
(#1965). Neither reproduces today. Both rest on a convention that nothing enforces, and both
fail loudly rather than quietly once they do happen.

Also audited and found genuinely safe, so left alone: `recentSearchStopsList` keys stops on a
bare `stopId`, which would collide with the results list, except the two live in
mutually exclusive `when (listState)` branches and never coexist.

## Changes

**Search result keys are prefixed by type** (`SearchStopScreen.kt`). The results list mixes
`Stop`, `Trip` and `Address`, and keyed each on its bare id. Those ids come from unrelated
sources: stop ids from the GTFS stop table, trip ids raw from the routes proto, address ids
from `stop_finder`. Nothing makes the three spaces disjoint, so one shared value throws
`Key ... was already used` and takes the screen down. Route results only appear when the
query is exactly a bus route short name, which is also when a short numeric id is most likely
to be echoed by both sides. Now `"stop_"`, `"trip_"` and `"address_"`, matching the prefixing
`addressResultsSection` and `emptyStateStopsList` already use.

Within-type duplicates were already impossible and are unchanged: `NswBusTripOptions.tripId`
is `TEXT PRIMARY KEY` with no join in `selectTripsByRouteIds`, and stops come back
`GROUP BY s.stopId`.

**Product class uniqueness is now guarded** (test only, no production change).
`TransportMode.byProductClass` is built with `associateBy`, so a duplicate `productClass`
does not fail, it silently overwrites: eight modes become a seven-entry map and
`fromProductClass` returns the wrong mode for the loser. That function is how every TfNSW
response is turned into a mode, so the damage is quiet and app-wide (wrong icon, colour and
name on journey cards, mode filters selecting the wrong chip, product-class analytics
attributed to the wrong mode) and the mode filter `LazyRow` in `TimeTableScreen` crashes on
the duplicate key.

This is a live footgun rather than a theoretical one: `SchoolBus` already shares both its
colour and its `name` with `Bus` and differs only by product class, so adding a mode by
copying that block is one forgotten edit away from the collision.

## Testing

Both suites were confirmed to fail against the unguarded code rather than merely passing:

- `SearchResultKeyCollisionTest` (new, 3 cases) renders the real screen with a trip and a
  stop sharing an id, and an address and a stop sharing an id. Reverting the prefixes fails
  2 of 3 with `IllegalArgumentException: Key "700" was already used`.
- `TransportModeProductClassTest` (new, 4 cases) asserts distinct product classes, the
  `associateBy` round trip per mode, that an unknown class resolves to nothing, and that
  SchoolBus stays separable from Bus. Giving `OnDemand` Bus's product class fails 3 of 4,
  including `Duplicate productClass in TransportMode.all: [5] ... expected:<8> but was:<7>`.

No device QA: the search results path needs a stop and a bus route sharing an id, which
cannot be produced against live data. The Robolectric cases cover it deterministically
instead.

### DIFF
--- a/core/transport/src/commonTest/kotlin/xyz/ksharma/krail/core/transport/TransportModeProductClassTest.kt
+++ b/core/transport/src/commonTest/kotlin/xyz/ksharma/krail/core/transport/TransportModeProductClassTest.kt
@@ -1,0 +1,67 @@
+package xyz.ksharma.krail.core.transport
+
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertNotNull
+import kotlin.test.assertSame
+
+/**
+ * Guards the one invariant three separate features assume: every [TransportMode] owns a
+ * distinct [TransportMode.productClass].
+ *
+ * Nothing enforces it today. `TransportMode.byProductClass` is built with `associateBy`, so a
+ * duplicate class does not fail, it **silently overwrites** - eight modes become a seven-entry
+ * map and [TransportMode.fromProductClass] starts returning the wrong mode for the loser.
+ * That function is how every TfNSW response is turned into a mode, so the damage is quiet and
+ * app-wide: wrong icon, wrong colour and wrong name on journey cards, mode filters selecting
+ * the wrong chip, product-class analytics attributed to the wrong mode, and the mode-filter
+ * `LazyRow` in `TimeTableScreen` crashing on a duplicate key.
+ *
+ * This is a live footgun rather than a theoretical one: [TransportMode.SchoolBus] already
+ * shares both its colour and its `name` with [TransportMode.Bus] and differs only by product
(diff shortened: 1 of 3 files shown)


ksharma-xyz:
* **#1966** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1966?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1966?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1970, opened 2026-09-03
feat(search): mask digits in the query sent with search_stop_query

ksharma-xyz:
## What

Attaches the typed search query to every settled search instead of only to the zero-result carve-out, with every digit masked out of it on the device.

Masking happens client-side because `RealAnalytics` logs directly to Firebase: there is no stage after this one that could remove a house number from a value already sent.

An all-digit query is sent as typed. Route numbers and stop IDs are digits with no street beside them, so they identify no address, and masking them would collapse every such search into a run of `#`. One non-digit character anywhere masks the whole query.

The event also stops firing per keystroke: it now waits 600 ms of quiet after the results render, and the next keystroke cancels the job while it waits. A typing burst reports the query the rider finished rather than one row per prefix. The list on screen still refreshes at 100 ms.

## Changes

- `SearchQueryAnalyticsRedaction`: `zeroResultQueryOrNull` becomes `maskedQueryOrNull`. Result counts are no longer part of the decision; the length cap goes from 25 to 40 so real stop names fit ("north sydney interchange stand c" is 32 characters)
- `AnalyticsEvent.SearchStopQuery`: `zeroResultQuery` renamed `maskedQuery`. The Firebase property key stays `query`, so existing dashboards read the same column
- Only the local firing carries the text. The address firing carries none, since both share a `searchSessionId` and duplicating it would double the egress
- `resolveLocalZeroResultQuery` deleted, along with the two-pipeline ownership rule it existed for
- `SearchStopViewModel`: analytics settle time split from the list debounce
- Fixes a defect in the same path: `runCatching` swallowed the `CancellationException` thrown when a keystroke cancelled the search job, so a cancelled keystroke rendered the error state and reported `isError = true`. Replaced with `suspendSafeResult`, which rethrows cancellation
- `AnalyticsParamSanitizer` untouched: address-ID hashing and param truncation are unrelated

Guard tests re-pinned to the new invariant, which is simpler than the carve-out it replaces: a digit only ever leaves as part of a query that is nothing but digits. Adds a test pinning the burst rule (three keystrokes, one reported search).

## Testing

- `./gradlew :feature:trip-planner:ui:testAndroidHostTest :core:analytics:testAndroidHostTest --continue` green
- `./scripts/fullQualityChecks.sh` green (Android compile, iOS simulator compile, detekt)
- Not yet exercised on a device; DebugView verification is tracked in `docs/SEARCH_TELEMETRY_1.27_TODO.md`

### DIFF
--- a/core/analytics/src/commonMain/kotlin/xyz/ksharma/krail/core/analytics/event/AnalyticsEvent.kt
+++ b/core/analytics/src/commonMain/kotlin/xyz/ksharma/krail/core/analytics/event/AnalyticsEvent.kt
@@ -230,23 +230,22 @@ sealed class AnalyticsEvent(val name: String, rawProperties: Map<String, Any>? =
     }
 
     /**
-     * Fired when a settled search query finishes resolving. Never carries the raw
-     * typed text by default: a query can be a street address, which identifies a home
-     * or workplace, and the privacy policy promises analytics hold no personally
-     * identifiable data.
+     * Fired when a settled search query finishes resolving.
      *
-     * [zeroResultQuery] is the single, deliberately narrow exception, kept for
-     * fuzzy-matcher diagnostics (the "townhall returned nothing" workflow). Callers
-     * must resolve it through `SearchQueryAnalyticsRedaction.zeroResultQueryOrNull`,
-     * which requires zero results everywhere, no digits, and a short length. It lands
-     * in the same "query" property historical dashboards already read.
+     * [maskedQuery] carries what the rider typed, with every digit replaced by `#`. The
+     * privacy policy discloses the collection and promises the masking; callers must
+     * resolve it through `SearchQueryAnalyticsRedaction.maskedQueryOrNull` rather than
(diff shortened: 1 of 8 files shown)


ksharma-xyz:
* **#1972** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1972?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1971** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1971?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1970** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1970?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1970?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1974, opened 2026-09-05
docs(search): correct the address gate finding to a measurement artifact

ksharma-xyz:
## What

`docs/SEARCH_AND_LABELS_TELEMETRY_FINDINGS.md` on `main` files the `BELOW_THRESHOLD` digit-bearing bucket as the one user-visible defect, at P1, and proposes lowering the address search length threshold. That reading was wrong, and the doc is live, so anyone acting on it would change the gate to solve a problem that mostly is not there.

`search_stop_query` used to fire once per settled keystroke rather than once per search. The local debounce is 100 ms, so typing slower than that emitted a firing per character. Every prefix of `4 fulton place` is short enough to be `BELOW_THRESHOLD`, contains a digit, and finds nothing locally, so each one landed in the bucket as an independent failed search while the rider's actual search succeeded moments later. Nothing in the payload distinguished a keystroke from an abandoned search, so the two could not be told apart from the data alone.

Most of that bucket was riders mid-typing. What remains is not clearly addresses either: a short all-digit query is more likely a route number or stop ID, which an address lookup answers no better than the local stop search does.

## Changes

- Rewrite item 1 as a corrected artifact rather than deleting it. The bucket is still in the data and still looks compelling to anyone querying it cold, so the section now carries the correction to stop the same conclusion being re-derived
- Update the summary table verdict and the "only item 1 is a defect users are feeling today" claim, and add a pointer at the top of the doc
- Add `address-gate-digit-queries` to `docs/ANALYTICS_ASSUMPTIONS.md`, so the claim has a scheduled re-check instead of living only in a handover doc. That it never entered that ledger is the process half of the failure: the ledger exists for exactly this, and a data-derived claim reached a P1 label without passing through it
- Add `docs/learning/2026-09-05-a-bucket-that-was-mostly-typing.md`

## The audit

The learning entry records an audit of every analytics call site reachable from a debounce or a repeating loop, since the generalizable defect is an event that can fire more than once per user intent without carrying anything that identifies the intent. Four candidates besides the one already fixed, all already intent-shaped:

- `DeparturesViewModel` fires its status event only on the first transition into error per polling session
- `TimeTableViewModel`'s auto-refresh loop pokes the rate limiter and tracks nothing
- `DeviceWindowReporter` debounces, `distinctUntilChanged`s, and compares against the last *reported* snapshot
- `SavedTripsViewModel` fires from `onEvent` handlers; its one delayed path arms a review moment and sends no analytics

## Intentionally not changed

`AddressSearchEligibility`. The threshold stays where it is.

## Testing

- `scripts/check_stale_assumptions.py` parses the new ledger row (4 tracked)


ksharma-xyz:
* **#1974** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1974?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1974?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1983, opened 2026-09-11
feat(connectivity): implement the observer on Android and iOS

ksharma-xyz:
## What

Implements `ConnectivityObserver` on both platforms behind the contract added in #1982, wires the Koin modules, and adds the test fake. Still inert: nothing injects `ConnectivityObserver` yet, so no runtime behaviour changes.

## Changes

- `AndroidConnectivityObserver` over `ConnectivityManager.registerDefaultNetworkCallback`. Gates on `NET_CAPABILITY_VALIDATED`, not `NET_CAPABILITY_INTERNET`. The latter is true the moment a phone joins a captive-portal Wi-Fi, before anything can actually be reached, so gating on it would make this class a confident source of wrong answers, which is worse than having no source at all. `VALIDATED` means the platform probed and got through.
- The Android flow seeds itself from the current network before registering the callback. Without that it sits at `Unknown` until something about the network changes, which on a device connected for hours may be never.
- `IosConnectivityObserver` over `nw_path_monitor` from the Network framework, which Kotlin/Native exposes without a cinterop def file. No dependency taken: this is roughly sixty lines over an API already available, against a library that would have to be read anyway the first time it misbehaved. It has no equivalent of `VALIDATED`, so a captive portal reads as satisfied there. That gap is closed later by the failure classifier, not here.
- Both `StateFlow`s are shared `Eagerly` and both Koin singles are `createdAtStart`. A lazily created observer would be constructed at the moment of the first failed request and report `Unknown` for exactly the classification that needed it. A subscriber-gated share has the same defect.
- `logTransitions()` is the only runtime signal these produce. The platform classes are thin wrappers over OS callbacks that no unit test can reach, so toggling airplane mode and reading the log is how they get verified at all.
- `ACCESS_NETWORK_STATE` added to `androidApp/src/main/AndroidManifest.xml`. Normal install-time permission, no runtime prompt.
- `FakeConnectivityObserver` in `:core:testing`, exposing a `MutableStateFlow` so a test can drive offline and online transitions directly.
- `connectivityModule` registered in `KoinApp.kt`.

## Testing

- `./gradlew :core:connectivity:testAndroidHostTest` green
- `./gradlew :androidApp:assembleDebug` green
- `./gradlew :core:connectivity:compileKotlinIosSimulatorArm64` green
- `./scripts/fullQualityChecks.sh` green
- detekt green, no autocorrected files

Android emulator, API 36:

| Step | Observed |
|---|---|
[...]

PR #1982, opened 2026-09-11
feat(connectivity): add TransportState and the observer contract

ksharma-xyz:
## What

Adds `:core:connectivity` as a contract only: the `TransportState` type, the `ConnectivityObserver` interface, the `reconnections()` operator and their tests. No platform implementation and no consumer, so nothing in the app behaves differently yet.

First of a stack that gives the app a typed answer to "why did that request fail" instead of a single `isError` boolean.

## Changes

- `TransportState`: three states rather than a `Boolean`. `Unknown` is the initial value and carries weight. Reporting `Down` before the first OS callback would make a perfectly good first request fail fast, and would let the app claim a rider is offline on no evidence. `shouldAttemptRequest` treats `Unknown` as permission to try; `mayClaimOffline` treats it as grounds not to claim anything.
- The type is named for the transport rather than `isOnline` on purpose. The OS reports the state of the local transport, not whether anything can be reached: a phone behind a captive portal, on a dropped VPN tunnel, or on a carrier blackholing packets all report a usable network. Transport state is an input to classifying a failure; the request outcome is the authority.
- `ConnectivityObserver`: a hot, app-scoped `StateFlow<TransportState>`, readable synchronously at the moment a request throws.
- `reconnections()`: emits once per transition into connected. All three operators are load-bearing. `distinctUntilChanged` collapses the repeat callbacks Android fires while a network settles, `drop(1)` stops a collector seeing a reconnection that never happened on every screen entry, and `filter` keeps only the transitions in.
- New Gradle module with `withHostTest {}`, wired into `settings.gradle.kts`.

Deliberately not in this PR: the platform implementations, DI, the manifest permission and the test fake. Those are the child branch. An `expect`/`actual` pair cannot be split across PRs, and this half compiles and tests standalone without one.

## Testing

- `./gradlew :core:connectivity:testAndroidHostTest` green, 8 tests
- `./gradlew :core:connectivity:compileKotlinIosSimulatorArm64` green
- `./scripts/fullQualityChecks.sh` green
- detekt green, no autocorrected files

Each `reconnections()` test covers a case that has produced a real bug in designs of this shape: a refetch firing on every screen entry because the current value was not dropped, a burst of refetches because settling callbacks were not collapsed, and an offline claim made before the first callback because `Unknown` was folded into `Down`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01SKyPckiZry4wKSq9W9TLSL

PR #1984, opened 2026-09-11
feat(network): add the NetworkError taxonomy and NetworkException

ksharma-xyz:
## What

Adds `NetworkError`, a typed vocabulary for why a network call failed, plus `NetworkException` as the single throwable services put into a failed `Result`. Nothing produces one yet: the classifier is #1985 and the service migration is later in the stack.

Until now every network failure reached the UI as `isError: Boolean`, so airplane mode, an NSW outage and a changed response shape were indistinguishable and produced the same screen.

## Changes

- `NetworkError`: eight cases. The two easiest to confuse are deliberately separate. `Offline` means the device has no transport, which is the rider's connection. `Unreachable` means transport was fine but the host could not be reached, which is not their fault and is worth retrying. Neither can be told apart from the thrown exception alone, which is why the classifier in #1985 takes the observed transport state as a second input.
- `isRetryable`, `recoversOnReconnect` and `isConnectionProblem` live beside the type rather than at call sites, and each is an exhaustive `when` with no `else`. A new case cannot compile until someone has decided what it means for all three. Two screens can no longer disagree about whether an NSW outage is the rider's fault.
- `Offline` is deliberately not retryable even though it obviously can succeed later. Retrying immediately just burns the connect timeout again, so it recovers through the reconnect path instead.
- `kind` is written out rather than derived from the class name, so renaming a case cannot silently split an analytics metric in two. It ignores the status code so upstream failures aggregate into one bucket.
- `NetworkException` keeps `kotlin.Result` rather than introducing a sealed `ApiResult`. A sealed type makes failure impossible to ignore, which is better in the abstract, but rewrites every service signature, call site and fake in a codebase where `suspendSafeResult` already threads `Result` and already handles `CancellationException` correctly. The looseness accepted is that `Result` allows a failure to be dropped silently; the trigger for revisiting that is recorded in the doc later in the stack.
- `asNetworkError()` falls back to `Unknown` rather than throwing. A call site handling a failure is the worst possible place to raise a second one.

## Testing

- `./gradlew :core:network:testAndroidHostTest` green, 12 new tests
- `./gradlew :core:network:compileKotlinIosSimulatorArm64` green
- detekt green

The tests pin the decisions the compiler cannot check: that a 4xx is never retried, that `Offline` is the only case recovery waits for, and that an upstream outage is never presented as the rider's connection problem.


ksharma-xyz:
* **#1984** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1984?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1984?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1985, opened 2026-09-11
feat(network): classify throwables into NetworkError on both platforms

ksharma-xyz:
## What

Implements `toNetworkError`, the `expect`/`actual` that turns a thrown exception plus the transport state observed when it landed into a typed `NetworkError`. Nothing calls it yet; the services start returning it later in the stack.

## Changes

- The signature takes two inputs, and that is the crux of the design. Neither source can classify alone. The OS transport state is a hint rather than a fact, because a device behind a captive portal, on a dropped VPN or on a carrier blackholing packets all report a usable network. And the exception alone cannot do it either: on Android, airplane mode and a genuine DNS failure on a working network both surface as `UnknownHostException`. The same throwable therefore classifies as `Offline` or `Unreachable` depending on what the OS was reporting, which are different sentences to a rider and only one of them is their fault.
- Note the parameter polarity: `isTransportDown`, matching `TransportState.mayClaimOffline`, which stays false while the state is still `Unknown`. An app that has not yet heard from the OS must not tell a rider they are offline.
- Everything Ktor raises above the engine is classified once in `toSharedNetworkError` and delegated to from both actuals, so the shared behaviour cannot drift between platforms. That drift has bitten this codebase before, see `docs/learning/2026-08-22-two-platforms-two-vocabularies.md`.
- `CancellationException` is rethrown rather than classified. A rider leaving a screen cancels the job, and turning that into a failure paints an error state over a screen nobody is looking at.
- `CaptivePortalException` exists so a Wi-Fi login page can be told apart from a genuine schema change. Both surface as a failure to read the body and only the content type separates them, so the distinction has to be drawn while the response headers are still in hand. The validator that raises it lands with the client factory work.

## Testing

- `./gradlew :core:network:testAndroidHostTest` green, 10 new tests
- `./gradlew :core:network:compileKotlinIosSimulatorArm64` green
- detekt green

The Android tests drive a **real** Ktor OkHttp client at real addresses rather than constructing exception instances by hand: an unresolvable `.invalid` host (RFC 2606), an unrouted TEST-NET-3 address (RFC 5737) and a closed loopback port. A test that hand-builds `UnknownHostException` proves the branch is wired and proves nothing about whether OkHttp throws that type for the condition in question. If an engine upgrade changes the exception surface, this fails instead of silently reclassifying live failures.

None of those addresses leave the machine in a way that reaches a third party, and none depend on the network being in any particular state, so this runs the same on CI as locally.

**The iOS mappings are not verified.** They are written from Apple's documented `NSURLErrorDomain` codes and have not been exercised against a real network change, because an iOS simulator shares the host's network stack and has no airplane mode, so the no-transport case cannot be produced there at all. The file says so at the top. The codes should be checked against observed `origin.code` values the first time this runs on a device.


ksharma-xyz:
* **#1985** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1985?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1985?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1987, opened 2026-09-11
fix(scripts): run the four CI verify tasks in fullQualityChecks

ksharma-xyz:
## What

`scripts/fullQualityChecks.sh` is the documented pre-PR gate, but it ran none of the structural verification tasks that `code-quality.yml` fails on. A branch could be green locally and still fail CI on a task the script never invoked.

That is not hypothetical. A new module with `commonTest` sources and no iOS lane classification passed layout invariants, both compiles and detekt locally, then failed `code-quality / detekt` on `verifyIosTestClassification`.

## Changes

- Runs the same four tasks as the workflow's "Verify test wiring" step: `verifyTestWiring`, `verifyTestingModuleUsage`, `verifyNoAdHocBoundaryFakes`, `verifyIosTestClassification`.
- They run **before** the two compiles, not after. Each is a seconds-long structural check, so a module missing `withHostTest {}` or missing a lane classification now fails in seconds rather than after a full Android and iOS compile.
- `--continue` matches CI, so all four report at once rather than one failure per run.
- `-PciQuality` is deliberately not passed. That flag exists so CI can substitute placeholder API keys; a local run has the real ones in `local.properties`.
- The header comment now states the coupling: this script has to cover everything CI can fail on short of a full build, and a verification task added to the workflow gets added here in the same change. That coupling is the actual fix. The four task names are just its current contents.

Left alone on purpose: `check_stale_assumptions.py` stays a warning, and the two compiles keep their order. This PR only adds the missing step.

## Testing

- `./scripts/fullQualityChecks.sh` green on a clean `main`, with the new step reporting between the Python checks and the Android compile:

```
▶ Layout invariants...
▶ Analytics assumptions...
▶ Test wiring and lane classification...
▶ Android compile...
▶ iOS compile (Simulator arm64)...
▶ Detekt...
✓ All quality checks passed.
```

- The failure path is already evidenced: `verifyIosTestClassification` failed CI on an unclassified module while every other local check was green, which is the gap this closes. `set -e` plus Gradle's non-zero exit under `--continue` aborts the script, the same property the existing detekt step relies on.


ksharma-xyz:
* **#1987** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1987?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1987?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1992, opened 2026-09-11
feat(analytics): record network failures and recoveries as one event

ksharma-xyz:
## What

One event name for network failures and recoveries, not one per surface and not one per error kind, following the aggregation pattern in `docs/ANALYTICS_EVENTS.md`. The 500-name budget does not survive a name per failure mode.

Budget count in that doc updated (52 to 53), and a `Pending` row added to `docs/ANALYTICS_REGISTRY_HANDOFF.md` in this PR as `CLAUDE.md` requires.

## Changes

- `NetworkStatusEvent(action, errorKind, transportUp, upstream, endpoint)`, with `action` being `failure`, `recovered` or `retry`.
- **`transportUp` is the parameter worth having.** It is the classifier checking its own work: a run of `errorKind=offline` rows carrying `transportUp=true` means the classifier is calling a working connection offline. A classifier that is confidently wrong is worse than the single boolean it replaced, and nothing else in the app can detect that drift.
- Fired from `NetworkCaller` rather than from each ViewModel, because "once per transition into the failed state" needs somewhere that can see the transition. An offline device on a polling screen calls the same endpoint every 30 seconds; without the failed-endpoint set this would fire twice a minute, per screen, for as long as the rider stays in the tunnel. A success on an endpoint that was failing fires `RECOVERED` and clears it.
- Each service passes a **stable endpoint label** rather than the URL it hit. A trip can go to NSW direct, to the BFF's JSON pass-through, or to the BFF's proto endpoint, and labelling by the real path would split one metric three ways. The `upstream` param already carries that distinction. Labels are paths only, never query strings, because query strings carry stop ids.
- Resolving the base URL moved out of the classified block in all four services. It reads the debug store and a Remote Config flag, not the network, so a failure there was never a network failure to classify; it also has to happen before the call so the upstream label is known.

## Testing

- `./scripts/fullQualityChecks.sh` green
- `./gradlew testAndroidHostTest --continue` green
- `./gradlew iosUnitTest` green
- detekt green

Three unit tests cover the dedupe directly: repeated failures on one endpoint fire once, a success after a failure fires `RECOVERED` exactly once, and a success with no prior failure fires nothing.

### Device, Android emulator, timetable open, Town Hall to Central

```
15:06:03  network call failed: offline (transportDown=true)
15:06:03  NetworkStatusEvent(action=FAILURE, errorKind=offline,
          transportUp=false, upstream=NSW, endpoint=/v1/tp/trip)
15:06:34  transport=Up, refetching
[...]

PR #1991, opened 2026-09-11
feat(network): gate polling, decline retries and refetch on reconnect

ksharma-xyz:
## What

The phase where riders notice something. Everything before this in the stack shipped inert.

## Changes

- **State carries the error.** `TimeTableState` and `DeparturesState` gain a `NetworkError` alongside the existing `isError` boolean. `isError` still decides whether to show an error surface; the new field says what that surface should say. The UI work is tracked separately, so this only widens the state so that work is purely additive.
- `NetworkError` is marked `@Immutable` because a screen's state holds one. Every case is a data object or holds a single `Int`; `Unknown`'s `cause` is written to logs and never read by rendering code.
- **Refetch on reconnect.** `TimeTableViewModel` refetches when transport returns after an offline failure. Only `NetworkError.Offline` arms it: an NSW outage is not fixed by reconnecting, because the connection was never the problem. `reconnections()` drops the current value, so a configuration change cannot fire a spurious refetch on resubscription.
- **Polling gate.** `DepartureBoardRepository`'s poll loop suspends while the OS reports no transport. Polling a stop every 30 seconds from inside a tunnel spends battery to collect a connect timeout. `TransportState.Unknown` counts as connected: we have not heard from the OS yet and must not refuse to try on no evidence.
- **Retry.** `HttpRequestRetry` installed on the base client so a future upstream inherits it rather than opting in. It declines entirely while transport is down, never retries a 4xx, and caps at two attempts because polling screens already refetch every 30 seconds and a retrying poll doubles load on NSW during exactly the outage that caused the retry. Backoff is jittered so a stop many riders are watching does not produce a synchronised burst when NSW recovers.

## Testing

- `./scripts/fullQualityChecks.sh` green
- `./gradlew testAndroidHostTest --continue` green
- `./gradlew iosUnitTest` green
- detekt green

`NetworkCallerTest` covers the classification path, including the case that motivates the whole design: transport dropping **mid-flight**, where a caller that captured the state up front would get the answer wrong.

### Device, Android emulator, Central to Town Hall with the timetable open

| Time | Observed |
|---|---|
| 14:55:41 | `NSW GET /v1/tp/trip` — online, loads |
| 14:55:55 | `transport=Down` — airplane mode on |
| 14:56:12 | `NSW GET /v1/tp/trip` — 30s auto-refresh |
| 14:56:12 | `network call failed: offline (transportDown=true)` — **+23ms** |
| 14:56:24 | `transport=Up` — airplane mode off |
[...]

PR #1990, opened 2026-09-11
refactor(network): type park-ride failures and fill the service register

ksharma-xyz:
## What

Completes the contract unification and fills in the service register.

`RealParkRideService` already returned a `Result`, but its failures were raw throwables from `suspendSafeResult`, so a caller could tell that something failed and nothing about what. It now goes through `NetworkCaller` like the other two.

## Changes

- `fetchCarParkFacilities` (both overloads) return typed failures.
- `fetchAvailabilityForStops` keeps throwing rather than returning a `Result`. Its `null` already means "BFF off, use the per-facility path", and a `Result<T?>` would give callers two ways to say nothing. It still routes through `NetworkCaller`, so what escapes is a `NetworkException`.
- The service register in `docs/NETWORK_RELIABILITY.md` is filled in, and `NetworkServiceRegisterTest` holds it: it reads the doc, scans for `Real*Service` classes, and fails when they disagree. It also fails on a row naming a service that no longer exists, because a register listing things that are gone is how it stops being read.

### What writing the guard found

It immediately surfaced `RealNswGtfsService` in `:gtfs-static`, which this stack had missed entirely. It is recorded as **not migrated**, with the reason: it downloads static GTFS archives at app start rather than serving rider-facing requests, nothing renders a message when it fails, and a retry is the next app start. Being caught by the guard rather than by somebody noticing is the register working.

`RealGtfsRealtimeService` stays on its own `GtfsRealtimeResult` sealed class, recorded as a deliberate exception. It has an `Unchanged` case that `Result<T>` cannot express, and its failures are consumed by a poller that falls back to direct polling rather than surfacing them, so folding it in would lose the third case for no gain.

## Testing

- `./scripts/fullQualityChecks.sh` green
- `./gradlew testAndroidHostTest --continue` green
- `verifyTestWiring verifyTestingModuleUsage verifyNoAdHocBoundaryFakes verifyIosTestClassification` green
- detekt green

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01SKyPckiZry4wKSq9W9TLSL

PR #1989, opened 2026-09-11
refactor(network): return typed Results from trip planning and departures

ksharma-xyz:
## What

`RealTripPlanningService` and `RealDeparturesService` threw; `RealParkRideService` returned a `Result`. Same layer, two contracts, so a typed error could only ever reach half the app. These two now return `Result<T>` whose failure is always a `NetworkException` carrying a `NetworkError`.

## Changes

- **`NetworkCaller`** is the single place classification happens. A service that reached for the `ConnectivityObserver` itself would be four services each deciding separately how to read it, which is how the app came to have two contracts in the first place.
- It reads the transport state **at the moment the call fails** rather than taking it as a parameter. A caller that captured it before the request would classify against a state up to 30 seconds stale, which is exactly long enough to cross a tunnel.
- `ensureActive()` runs before classification, so a cancelled job stays cancelled instead of becoming a failed `Result`.
- Most call sites got **smaller**, because they were already wrapping these calls to get a `Result`. `TimeTableViewModel.loadTrip` had a `runCatching` that built `Result.success` inside and unwrapped with `getOrElse` to rebuild `Result.failure` outside; that is now just the call. Another site did `runCatching { ... }.mapCatching { it }`. `DepartureBoardRepository` dropped its `suspendSafeResult` wrapper and with it an `ioDispatcher` constructor parameter it no longer used.
- Two sites keep a throwing shape on purpose, with `getOrThrow`: `TripPoller` and `RealRemoteAddressResultsManager` both already had a surrounding handler, and the throwable they now see is a `NetworkException` rather than a raw Ktor one.
- The fakes gained `failWith: NetworkError?` so a test can drive offline versus upstream behaviour rather than only "something went wrong".

## Testing

- `./scripts/fullQualityChecks.sh` green
- `./gradlew testAndroidHostTest --continue` green
- detekt green

### One change that is not cosmetic

Both fakes now `yield()` before returning. The real service always crosses a dispatcher inside `NetworkCaller`, so it always suspends. A fake that returned without suspending let a caller's loading state and its result coalesce into one `StateFlow` emission, which broke three `DepartureBoardRepositoryTest` cases that were correctly asserting the loading state is observable.

Those three failures were the fake hiding a conflation bug rather than reproducing production behaviour, so the fix was to make the fake faithful rather than to weaken the assertions.

Two ad-hoc `TripPlanningService` fakes in `:feature:track:ui` tests were updated in place rather than replaced with the canonical one; consolidating them is a separate change.

[...]

PR #1993, opened 2026-09-11
feat(network): detect captive portals so the case is reachable

ksharma-xyz:
## What

`NetworkError.CaptivePortal` was defined, classified and documented, and **nothing could produce it**. The response validator that raises `CaptivePortalException` was described in #1985 as landing "with the client factory work" and then never landed. The case was dead code.

## Changes

Portals arrive two ways, and only one of them reaches a validator.

**HTML body with a 2xx status** is the common shape. `expectSuccess` is satisfied, so nothing objects until the body fails to parse, and by then the headers are gone and a portal is indistinguishable from a schema change. `installCaptivePortalValidator` catches it while the `Content-Type` still exists.

It checks for HTML specifically rather than "not the type we expected". The stricter rule would also fire on an upstream that legitimately changed its type, on a `text/plain` error body, and on any endpoint added later with a different shape, turning a schema change into a confident "sign in to this Wi-Fi". Portals are HTML and nothing KRAIL calls serves HTML, so keeping the signal narrow is what stops false accusations.

**HTTP 511** is the second shape, handled in the classifier rather than the validator. RFC 6585 defines it as "the client needs to authenticate to gain network access", so it is a portal by definition, but `expectSuccess` throws on non-2xx **before** `validateResponse` runs and the validator never sees it. Without that branch a 511 lands in the 5xx range and is reported as an NSW outage, telling a rider the transport network is down when they only need to join a Wi-Fi.

That second case was found by a failing test rather than by reading the code. The test asserting "HTML is a portal regardless of status code" failed against a real Ktor pipeline, which is the argument for driving these through a client rather than calling the validator directly.

Taxonomy table in `docs/NETWORK_RELIABILITY.md` updated with both rows and the reasoning.

## Testing

- `./gradlew :core:network:testAndroidHostTest` green, 6 new tests (37 across the module)
- detekt green

The six tests cover both portal shapes, JSON and protobuf passing through untouched so the BFF proto endpoints are not mistaken for portals, a genuine 5xx staying an upstream failure, and the derived properties (never retried, is a connection problem, does not recover on reconnect).

Not covered: a real café portal. That needs hardware or a local proxy and is on the testing plan rather than in this PR.


ksharma-xyz:
* **#1993** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1993?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1993?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1988, opened 2026-09-11
refactor(network): replace five per-feature HTTP clients with one factory

ksharma-xyz:
## What

Four feature modules each carried an `expect fun <feature>HttpClient(baseClient)` with two `actual` implementations. Twelve files, every one byte-for-byte identical apart from the function name and which BuildKonfig constant it read. The only real difference between a "trip planner client" and a "departures client" was nothing.

Nobody added them carelessly. Each was a reasonable local decision, copied from the module next door, and the duplication was only visible from above.

## Changes

- Twelve files become two: an `ApiCredential` enum and a `forApi` extension in `:core:network`, plus one `expect val nswTransportApiKey` for the only genuine platform difference. The branch is over a string now, not over a client factory.
- A new upstream costs an enum case, not three files.
- `NetworkClientFactoryGuardTest` holds both halves of the mistake: no function named `*HttpClient` declared outside `:core:network`, and no module outside `:core:network` reading an NSW API key, since a module reaching for the key is a module about to build its own client.

Pure refactor. `forApi` appends exactly the header the deleted functions appended, Ktor's `config()` returns a new client over the same engine, and every caller keeps the timeouts, logging and content negotiation from the base client. No runtime behaviour changes.

Anything that should apply to every KRAIL request now has one obvious home on the base client, which is what the retry plugin and the response validator need later in the stack.

## Testing

- `./scripts/fullQualityChecks.sh` green
- `./gradlew testAndroidHostTest --continue` green
- `verifyTestWiring verifyTestingModuleUsage verifyNoAdHocBoundaryFakes verifyIosTestClassification` green
- detekt green

The guard was verified by reintroducing an offender rather than trusting that it passes: with a `departuresHttpClient` function put back, it fails and names the file and line. Removed again before commit.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01SKyPckiZry4wKSq9W9TLSL

PR #1986, opened 2026-09-11
docs(network): record the failure taxonomy and the decisions behind it

ksharma-xyz:
## What

Adds `docs/NETWORK_RELIABILITY.md` and points `CLAUDE.md` at it. The proposal that argued for this work stops being true once the work lands; this is what survives it.

## Changes

- **The taxonomy table** is the specification for `toNetworkError` and the fixture its tests are written from, so it lives in one place rather than being copied into the classifier, the tests and eventually the UI copy. Three consumers, one table.
- **A verification column that says plainly the iOS half is a hypothesis.** Those mappings come from Apple's documented `NSURLErrorDomain` codes and have never been exercised against a real network change, because an iOS simulator shares the host's network stack and has no airplane mode. The Android half is verified against a real Ktor OkHttp client. Recording that asymmetry is the point: a classifier that is confidently wrong is worse than the boolean it replaced.
- **A decisions table** borrowing the column that makes `docs/ANALYTICS_ASSUMPTIONS.md` work: "would revisit if". Not "consequences", which nobody rereads, but the specific observation that should reopen the question. All five triggers are concretely observable, including the one that settles the `Result` versus `ApiResult` argument with a single real instance of a silently dropped failure.
- **A checklist for adding an upstream API**, and the two absolute rules in `CLAUDE.md`: never add a per-feature `HttpClient`, never present the OS connectivity flag as the truth.
- **The service register**, present but empty. It gets populated as the services migrate to one contract, along with the guard test that holds it. Writing the register before the contract exists would be writing fiction.

Deliberately not a `docs/adr/` directory. All five decisions are scoped to one subsystem, the rule docs here already carry their rationale, and a new genre seeded with five entries and no second customer is one that rots. The section lifts out into `docs/adr/0001` unchanged on the day a second subsystem needs the same treatment.

## Testing

- `./scripts/fullQualityChecks.sh` green
- `./gradlew testAndroidHostTest --continue` green
- detekt green

Docs only, no code changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01SKyPckiZry4wKSq9W9TLSL

PR #1994, opened 2026-09-11
feat(network): emit the retry event so retrying can be measured

ksharma-xyz:
## What

`NetworkStatusEvent.Action.RETRY` was defined in #1992 and **nothing fired it**. The plugin retries inside the Ktor pipeline where nothing else in the app can see it, so every retry was silent and the enum case was unreachable.

That matters because the reason the case exists is to answer whether retrying earns its cost. Comparing retry rows against failure rows is the only way to tell a policy that recovers requests from one that multiplies load on NSW during exactly the outage that caused the retry. Without the rows there is no way to know which one shipped.

## Changes

- `installKrailRetry` takes an observation hook; `NetworkCaller.recordRetry` turns it into the event.
- Retry rows are deliberately **not** deduped, unlike failures. The interesting number is how many retries happened, and the plugin caps at two per request so it cannot run away.
- The endpoint is read from the request inside the pipeline rather than passed down, because the service's stable label is not in scope there. Path segments only, never the query string, which carries stop ids.
- Koin resolves `NetworkCaller` lazily inside the client's lambda rather than capturing it, so the client and the caller do not depend on each other at construction time.

No new event name and no new params, so no registry ledger row: `action=retry` was already declared in #1992's row.

## Testing

- `./gradlew :core:network:testAndroidHostTest` green, 5 new tests
- `verifyTestWiring verifyTestingModuleUsage verifyNoAdHocBoundaryFakes verifyIosTestClassification` green
- detekt green

The five tests drive the real policy over a `MockEngine`: a 5xx retries and the hook fires with a path and an upstream, a 4xx never retries, retries cap at three total attempts, and **nothing retries while transport is down**.

That last one is what keeps an offline failure costing one connect timeout instead of three, and it is invisible to any test that does not actually run the plugin. It is also the property behind the 23ms offline failure measured on device in #1991.


ksharma-xyz:
* **#1994** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1994?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1994?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1999, opened 2026-09-12
docs(network): record the iOS offline classification verified on a device

ksharma-xyz:
## What

The iOS half of the failure taxonomy was written from Apple's documented `NSURLErrorDomain` codes and had **never been observed**, because a simulator shares the host's network stack and has no airplane mode. One row is now verified on an iPhone 11 running iOS 26.6.

## Changes

Airplane mode with the timetable open produced:

```
ERROR: network call failed: offline (transportDown=true)
DarwinHttpRequestException: Error Domain=NSURLErrorDomain Code=-1009
  "The Internet connection appears to be offline."
  _NSURLErrorNWPathKey=unsatisfied (No network route)
```

Three things at once:

- The code matches `NSURL_ERROR_NOT_CONNECTED_TO_INTERNET = -1009L` in the classifier
- The `Offline` branch was taken, not `Unreachable`
- `nw_path_monitor` independently agreed there was no route. **The iOS observer had never been exercised at all before this**, because a simulator cannot produce the state

Recovery verified in the same run: the failure count was sampled every 25 seconds for five minutes after airplane mode went off and never moved, so ten consecutive 30-second auto-refresh cycles completed cleanly.

### The verification row moves to "partly", not "yes"

Only the `Offline` row is observed. The `unreachable`, `timeout` and TLS rows still need a routeless Wi-Fi and remain hypotheses. Claiming the whole column was verified off one row would be exactly the confident-wrong-answer the taxonomy exists to prevent.

### Capture recipe added to the manual test doc

Working this out cost time and none of it is obvious:

- Kermit's debug lines do **not** reach stdout on a device. Only `logError` does, because `Log.ios.kt` calls `NSLog` explicitly. That is enough here, since the classification line and the underlying throwable are both at error level.
- `--terminate-existing` matters. Attaching to an already-running process captures nothing, which is how the first attempt produced an empty log.
- **Airplane mode cannot be toggled from the Mac.** iOS exposes no API and `devicectl` has no such command, so this case always needs a person holding the phone.

### One finding the run turned up

The timetable auto-refresh is **not** connectivity-gated, unlike the departure board's poll loop. It kept firing every 30 seconds while offline: two failures, 30 seconds apart. Each fails fast because the retry gate declines, so the cost is small, but the two surfaces disagree and `docs/POLLING_LIFECYCLE.md` now implies they do not.

Recorded as manual case 1.1.7 rather than fixed here, so it is not rediscovered. Worth its own change.

## Testing

- Device run: iPhone 11, iOS 26.6, 2026-09-12, build from `main` at the time of the run
- App terminated after the session so the timetable did not keep polling NSW

Docs only, no code changes.


ksharma-xyz:
* **#1999** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1999?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/1999?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #2002, opened 2026-09-19
chore(network): move Ktor to 3.6.0 and share one client configuration

ksharma-xyz:
## What

Moves the Ktor client from 3.4.1 to 3.6.0, and collapses the two platform copies of the
client configuration into one `commonMain` block behind an engine-only `expect`/`actual`.

No API KRAIL uses changed in 3.4.1 to 3.6.0, so nothing here is forced by the upgrade.

## Changes

**Upgrade**

- `gradle/libs.versions.toml`: `ktor = "3.4.1"` to `"3.6.0"`.
- The fixes in this range that land in KRAIL's hot path are all about cancellation, which
  the app does constantly because polling is gated on `WhileSubscribed`: KTOR-9773 (OkHttp
  could close a response body on the Android main thread when cancelled), KTOR-9762
  (Android could hang cancelling a streaming response), KTOR-9627 (blocking bridges cancel
  with the coroutine), KTOR-9870 (fewer OkHttp dispatches and allocations), KTOR-9834 (the
  client no longer initialises SLF4J eagerly during Android startup). On iOS, KTOR-9817
  fixes a Core Foundation over-release in Darwin's `CertificatePinner`.

**One configuration instead of two**

`baseHttpClient` was an `expect`/`actual` pair whose two bodies were a byte-for-byte copy
apart from the engine constructor. An edit to one was invisible to the other and nothing
compared them, so the first symptom would have been a behaviour that reproduces on one
platform only.

| File | Holds |
|---|---|
| `commonMain/HttpClient.kt` | `baseHttpClient()` |
| `commonMain/KrailHttpClientDefaults.kt` | `installKrailDefaults()`, every plugin |
| `commonMain/KrailHttpClientEngine.kt` | `expect fun krailHttpClientEngine()` |
| `androidMain/KrailHttpClientEngine.android.kt` | `OkHttp` |
| `iosMain/KrailHttpClientEngine.ios.kt` | `Darwin` |

Plugin order is unchanged, and is now documented where it is installed.

The engine stays named in source rather than resolved through Ktor 3.6's new
`ktor-client-engine-defaults` artifact. That artifact resolves to the right engines, but it
has no `androidJvm` variant (an Android consumer gets the JVM one, which also pulls in
`org.slf4j:slf4j-api`), and it converts an explicit decision into a resolution order. The
engine is not an implementation detail here: `error/NetworkErrorClassifier.android.kt` is
written against OkHttp's exceptions and `NetworkErrorClassifier.ios.kt` against Darwin's
`NSURLErrorDomain` codes.

**New test**

`ContentNegotiationAcceptHeaderTest`. Five services ask for `application/x-protobuf` and
then read the response with `readRawBytes()`. If `ContentNegotiation` appended
`application/json` to their `Accept` header, an upstream would be free to answer JSON to a
caller about to run a protobuf decoder. It does not append, measured with a `MockEngine` on
both 3.4.1 and 3.6.0, and the test pins that so a future upgrade fails the build instead of
a trip. Ktor 3.6 added `ContentTypeMergeStrategy.SkipIfPresent` (KTOR-5009) to make this
configurable; not set here, because the default already behaves this way.

**Dependencies removed**

- `ktor-client-auth` from six modules. No module installs the `Auth` plugin; credentials go
  on as a header from `ApiCredential`.
- The whole Ktor block from `:io:gtfs`, which imports no Ktor type.
- `ktor-client-darwin` from `:sandook`, a database module.

Left alone deliberately: the engine declarations in `:composeApp` and `:discover:ui`. They
look unused because no Kotlin file imports them, but both depend on `coil-network-ktor3`,
which builds its own client and needs an engine on the classpath.

## Testing

**Static**

- `./scripts/fullQualityChecks.sh` green (layout invariants, `compileDebugSources`,
  `compileKotlinIosSimulatorArm64`, `detekt`).
- `./gradlew testAndroidHostTest --continue` green across all modules.
- `./gradlew detekt --continue` green, no auto-corrected files left uncommitted.

**Android emulator (KRAIL_QA AVD)**

- `:androidApp:installDebug`, launched `xyz.ksharma.krail.debug`.
- Stop search renders results; Town Hall Station to Central Station returns live data.
  Logcat confirms the request goes out and comes back through the new shared pipeline:
  ```
  KrailNetwork: NSW GET /v1/tp/trip [override=on]
  KrailNetwork: RESPONSE: 200
  ```
- Rotated to landscape on the timetable screen with data loaded: dual pane renders, the
  refetch on recreation returns `200`. Rotated back, then navigated back twice.
- `adb logcat -d | grep -c "FATAL EXCEPTION"` is `0` for the whole session.

**iOS simulator (6.5", iOS 26.5)**

- `xcodebuild -scheme iosApp -configuration Debug` BUILD SUCCEEDED; installed and launched.
- Both Maestro lanes driven against the simulator with this build:

  | Lane | Flows | Result |
  |---|---|---|
  | `smoke/` | `01-launch-home`, `02-plan-trip` | 2/2 passed |
  | `nightly/` | `03-rotation-sweep`, `04-background-foreground`, `05-kill-relaunch`, `06-permissions-denied` | 4/4 passed |

- `02-plan-trip` is the one that matters here: it types a stop name, picks a result, does it
  again for the destination, then plans a trip and scrolls until a journey card is visible.
  All of that is live NSW traffic through the Darwin engine. The OS network log shows the
  request leaving the app:
  ```
  [com.apple.network:connection] Krail App ... url: https://api.transport.nsw.gov.au/v1/tp/stop_finder ... start
  ```
  and the timetable renders real multi-leg journeys with platforms and durations.
- `03-rotation-sweep` covers the configuration-change requirement on iOS, and
  `05-kill-relaunch` covers state restore.

**Android emulator, Maestro**

Same two lanes, same build, for symmetry with iOS:

| Lane | Result |
|---|---|
| `smoke/` | 2/2 passed |
| `nightly/` | 4/4 passed |

**Not run**

`:core:network:iosSimulatorArm64Test` cannot run at all: the test binary fails to link with
`ld: framework 'FirebaseCore' not found`. Pre-existing and unrelated to this change, and the
reason the repo does not run iOS unit tests. The Maestro lanes above cover the iOS path
instead.


ksharma-xyz:
* **#2003** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/2003?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2002** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/2002?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/2002?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ksharma-xyz:
## Device recordings

Both recordings are the same flow on this branch's build: cold launch, stop search for the origin, stop search for the destination, plan the trip, scroll to a journey card. Everything after the first tap is live NSW traffic through the shared client, so the video is the Ktor 3.6.0 request pipeline working end to end rather than a screen that merely rendered.

**Android emulator (OkHttp engine)**

https://github.com/user-attachments/assets/b06996ee-04b0-43b2-a527-3bdb07b00921

**iOS simulator, 6.5 inch, iOS 26.5 (Darwin engine)**

https://github.com/user-attachments/assets/ef144342-3990-4c8d-849b-b9ceb3cbb105

## Physical iPhone

Built, signed, installed and launched on a connected iPhone 11 running iOS 26.6. It starts and renders.

It was **not** driven end to end there, and the reason is tooling rather than the app: Maestro's bundled iOS device driver sets `IPHONEOS_DEPLOYMENT_TARGET` to 14.0, and the installed Xcode accepts 15.0 to 27.0, so the driver fails to build for a connected device:

```
error: The iOS deployment target 'IPHONEOS_DEPLOYMENT_TARGET' is set to 14.0,
but the range of supported deployment target versions is 15.0 to 27.0.x
```

Patching Maestro's own driver project to work around that would put the simulator lane at risk, which is the lane this repo actually runs, so it was left alone. The physical-device leg is install, launch and render only. Driving it needs either a Maestro release whose device driver targets 15.0 or a hand-run pass on the phone.

## Everything else

| Check | Result |
|---|---|
| `./scripts/fullQualityChecks.sh` | green |
| `./gradlew testAndroidHostTest --continue` | green, all modules |
| `./gradlew detekt --continue` | green |
| Maestro `smoke/` on iOS simulator | 2/2 passed |
| Maestro `nightly/` on iOS simulator | 4/4 passed |
| Maestro `smoke/` on Android emulator | 2/2 passed |
| Maestro `nightly/` on Android emulator | 4/4 passed |
| `adb logcat` FATAL EXCEPTION count | 0 |

`nightly/03-rotation-sweep` is what covers the configuration-change requirement on both platforms, and `05-kill-relaunch` covers state restore.

ksharma-xyz:
Physical-device gap now closed by hand.

The build from this branch was installed on a physical iPhone and exercised manually, and the app behaves correctly. That covers the one leg the automated run could not reach, for the reason given above: Maestro's connected-device driver targets `IPHONEOS_DEPLOYMENT_TARGET` 14.0 and the installed Xcode accepts 15.0 and up, so the driver never builds for a real phone. The simulator lanes and both Maestro suites were unaffected and are green.

So the Darwin engine on Ktor 3.6.0 is now verified on simulator (automated, recorded above) and on real hardware (manual).

PR #2003, opened 2026-09-19
docs(network): record the Ktor client architecture and version policy

ksharma-xyz:
## What

Adds `docs/KTOR_CLIENT_ARCHITECTURE.md`, the missing companion to
`docs/NETWORK_RELIABILITY.md`. That file covers what a failure means and what the app may
say about it; this one covers the client that produced it.

## Changes

- `docs/KTOR_CLIENT_ARCHITECTURE.md`:
  - The one-client rule and where each piece now lives after the engine seam.
  - Plugin order as a table, with the reason each plugin sits where it does (the two
    reorderings that would break something are named).
  - Why the engine is load-bearing for the error taxonomy, so it is declared in source.
  - The 3.6.0 upgrade record: which fixes it is actually for, and three things evaluated
    and **not** adopted with the measurement behind each: `ktor-client-engine-defaults`,
    `ContentTypeMergeStrategy.SkipIfPresent`, and the now-commonised `HttpCache`
    `FileCacheStorage`, which is tracked as follow-up rather than done here.
  - Which dependencies were removed and which look dead but are not.
  - A four-item list to work through the next time the version moves.
- `CLAUDE.md`: adds it to the per-feature doc list.
- `docs/NETWORK_RELIABILITY.md`: a pointer to it from the top.

Docs only. No source change in this PR.

## Testing

- `./scripts/fullQualityChecks.sh` green on the branch below this one.
- No code touched, so no test task applies to this PR on its own.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2004, opened 2026-09-20
docs(network): record the Ktor client architecture and version policy

ksharma-xyz:
## What

Adds `docs/KTOR_CLIENT_ARCHITECTURE.md`, the missing companion to
`docs/NETWORK_RELIABILITY.md`. That file covers what a failure means and what the app may
say about it; this one covers the client that produced it.

## Changes

- `docs/KTOR_CLIENT_ARCHITECTURE.md`:
  - The one-client rule and where each piece now lives after the engine seam.
  - Plugin order as a table, with the reason each plugin sits where it does (the two
    reorderings that would break something are named).
  - Why the engine is load-bearing for the error taxonomy, so it is declared in source.
  - The 3.6.0 upgrade record: which fixes it is actually for, and three things evaluated
    and **not** adopted with the measurement behind each: `ktor-client-engine-defaults`,
    `ContentTypeMergeStrategy.SkipIfPresent`, and the now-commonised `HttpCache`
    `FileCacheStorage`, which is tracked as follow-up rather than done here.
  - Which dependencies were removed and which look dead but are not.
  - A four-item list to work through the next time the version moves.
- `CLAUDE.md`: adds it to the per-feature doc list.
- `docs/NETWORK_RELIABILITY.md`: a pointer to it from the top.

Docs only. No source change in this PR.

Replaces #2003, which GitHub closed automatically when its base branch was deleted on merge of #2002. Same commit, rebased onto `main`.

## Testing

- `./scripts/fullQualityChecks.sh` green on the code change this documents, which merged as part of #2002.
- No code touched, so no test task applies to this PR on its own.


ksharma-xyz:
* **#2004** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/2004?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/2004?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #2005, opened 2026-09-20
fix(journeymap): drop a leg whose geometry cannot form a line

### LINKED ISSUE #1930: Journey map crashes when a leg yields exactly one coordinate
Found while writing the journey-map mapper tests for #1913 (PR #1923). Not fixed there, because that branch is test-only and baking a crash into an assertion would be the wrong record of it.

## What happens

`JourneyMapMapper` rejects a leg whose coordinate list is **empty**, but not one that yields exactly **one** valid coordinate pair. That single-point path reaches `JourneyMapFeatureMapper.createPathSegmentFeature`, which builds a `LineString` from it, and:

```
java.lang.IllegalArgumentException: LineString must contain at least two positions
```

Reproduced by a fixture with a one-point leg; the fixtures now use two points, so the path is currently uncovered rather than pinned.

## Why it can happen with real data

A leg's geometry comes from one of three sources (leg coords, interchange path, stop sequence). Any of them can degrade to a single locatable point: a leg with one usable coordinate, or a stop-connector where only one stop has a known position. `JourneyMapFeatureMapper` already drops single-locatable **connector** segments; the coords path has no equivalent guard.

## Fix shape

Decide per source, then guard in one place:

- [ ] Drop a path segment with fewer than two positions, the way the connector case already does, **or** promote it to a point feature if a single known position is worth drawing
- [ ] Move the check to where the geometry is built, so all three sources go through it
- [ ] Add the one-point fixture back and assert the chosen behaviour (it is the test that currently cannot exist)

## Severity

Crashes the journey map for that trip, so it is user-visible whenever the NSW response degrades that way. Not seen in the wild yet, and no reproduction from a real response is attached; the evidence is the fixture.

ksharma-xyz:
## What

Fixes #1930. A journey leg that yielded exactly one coordinate reached `LineString`, which throws `IllegalArgumentException: LineString must contain at least two positions` and takes the journey map down with it.

`JourneyMapFeatureMapper.createPathSegmentFeature` rejected an empty point list but not a single-point one. `createStopConnectorFeature` already had the right check, so the rule existed in one of the two places that needed it.

A leg can degrade to one locatable point from any of the three geometry sources: leg coordinates with one usable pair, an interchange path with one, or a stop sequence where only one stop has a known position.

## Changes

- `JourneyMapFeatureMapper.kt`: both feature builders now map their positions first and pass them through a single `lineStringOrNull` helper, which returns null below two positions. All three geometry sources are held to the same minimum instead of each one remembering it.
- The `validStops` pre-filter in `createStopConnectorFeature` is gone. `mapNotNull` over the stops already drops the unlocatable ones, and the size check now happens in one place for both builders.
- A dropped leg is logged with its leg id. Data degrading this way has not been seen in the wild, so the log is the only evidence we would get that it happens.
- `JourneyMapFeatureMapperTest.kt`: two tests. A one-point path leg is dropped rather than crashing, and a one-point leg sitting beside a drawable leg leaves the drawable one intact.
- `JourneyMapMapper` is untouched. It is free to emit a one-point `PathSegment`; the geometry rule belongs where the geometry is built.

## Testing

- `./gradlew :feature:trip-planner:ui:testAndroidHostTest` green, 13 tests in `JourneyMapFeatureMapperTest`
- The guard was verified by defeating it rather than by trusting it: with the minimum flipped to 1, three tests fail with `java.lang.IllegalArgumentException: LineString must contain at least two positions`. Restored before commit.
- `./scripts/fullQualityChecks.sh` green: layout invariants, analytics assumptions, `compileDebugSources`, `compileKotlinIosSimulatorArm64`, detekt


ksharma-xyz:
* **#2005** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/2005?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/2005?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #2007, opened 2026-09-20
fix(taj): draw the KRAIL wordmark in the rider's colour, unadapted

ksharma-xyz:
## What

The KRAIL wordmark in the Saved Trips title bar was drawn with `themeInkColor()`, which adapts the rider's chosen colour until it clears a contrast ratio against the surface behind it.

That is the right rule for a label someone has to read. It is the wrong rule for the brand. Nobody reads "KRAIL" to use the app, so the wordmark carries no contrast obligation, and adapting it made the one element that should be exactly the rider's colour the one element that was not.

It also disagreed with the same wordmark at the bottom of Settings, which was already drawing the raw colour. On any theme that needed adapting, the two rendered differently.

## Changes

- `SavedTripsScreen.kt`: the title bar wordmark now uses `themeDecorColor()`, the accessor whose stated purpose is recording that the unadapted colour is a decision rather than an oversight.
- `AppLogo.kt`: previously reached past the role vocabulary to `LocalThemeColor` and `hexToComposeColor` directly. It now uses the same accessor, so the two sites agree by construction rather than by coincidence. This is the change that makes the property hold on every theme, not just the one it was checked on.
- Nothing else moves. `themeInkColor()` remains correct for every label, icon and stroke that is actually read.

## Testing

- `./gradlew :feature:trip-planner:ui:detekt` green, including `ThemeColorRoleRule`
- `./gradlew :feature:trip-planner:ui:testAndroidHostTest` green
- Verified on device rather than by eye: the wordmark pixels on the Saved Trips title bar and at the bottom of Settings are both `#5AB031`, the raw theme green

## Snapshots

| Screen | Before | After |
|---|---|---|
| Saved Trips title bar | adapted green, darker than the rider's colour | `#5AB031`, the rider's colour exactly |
| Settings bottom logo | `#5AB031` already | `#5AB031`, unchanged, now via the named accessor |


ksharma-xyz:
* **#2007** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/2007?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/2007?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #2008, opened 2026-09-20
fix(taj): keep the alert chip bright and give it a derived rim

ksharma-xyz:
## What

Restores the alert chip fill to `#FFBA27` with a black label, and gives the chip its
edge from a derived outline rather than from the fill itself. The token had been
darkened to `#946500` to make the fill clear 4.5 against the card, which is the
contrast bar for text. A fill is not text, so chasing that bar produced a chip dark
enough that `getForegroundColor` flipped the label to near-white.

The substantive change is in the test. `KrailColorTokenContrastTest` was pairing
`alert` against `surface` at `TEXT_CONTRAST_AA`, and that assertion is what drove the
token dark. It now asserts what each element is actually held to: the outline against
the surface at the 3.0 non-text floor, and the label against the fill at 4.5.

## Changes

- `taj/theme/Color.kt`: `md_theme_light_alert` back to `#FFBA27`. `md_theme_dark_alert`
  unchanged at `#F4B400`.
- `taj/components/Button.kt`: adds `ButtonDefaults.alertOutlineColor(fill, isDark)`,
  which darkens the fill toward black on a light surface and returns it unchanged on a
  dark one. `AlertButton` draws it as a 1.5 dp border, alpha-matched to
  `LocalContentAlpha`.
- The rim is derived from the fill rather than stored as its own token, so it cannot
  drift from the colour it darkens, and adding a theme needs no new colour.
- `isDark` is passed rather than the surface colour because `KrailTheme.colors.surface`
  animates for 1500 ms during a theme switch, which would make the rim animate with it.
- `KrailColorTokenContrastTest.kt`: replaces the `alert` vs `surface` pairing with
  `alertOutline` vs `surface` at `UI_COMPONENT_CONTRAST_AA` and `onAlert` vs `alert` at
  `TEXT_CONTRAST_AA`. `DERIVED_NAMES` exempts both from the paired-or-excluded staleness
  guard, since neither is a `KrailColors` property.
- The border is applied only where the derived rim differs from the fill. On dark the
  derivation returns the fill, so painting it would add no edge and still redraw the
  rounded corner, which moved 330 antialiased pixels in the dark snapshot for nothing.
  Dark now renders byte-identical to before this branch.
- Records ten light-mode goldens: four `AlertButton` states in `taj`, and the three
  expanded `JourneyCard` previews in `trip-planner` that draw the chip in its real
  context. No dark golden changes.

The corrected test caught a real bug in the first version of the derivation:
`dark|alertOutline|surface` measured 2.23 against a 3.0 floor, because darkening a rim
on a dark surface reduces contrast. Hence the dark branch returning the fill unchanged.

## Testing

- `./gradlew :taj:verifyRoborazziAndroidHostTest :feature:trip-planner:ui:verifyRoborazziAndroidHostTest -PciQuality=true :taj:detekt :feature:trip-planner:ui:detekt --continue` green.
- `./gradlew :taj:testAndroidHostTest :feature:trip-planner:ui:testAndroidHostTest --continue` green.
- Installed on an emulator and checked the chip in light and dark.
- Rotated the screen and switched night mode on it. `adb logcat` reports no
  `FATAL EXCEPTION`.

## Snapshots

Pixel samples read off the device screenshot rather than thumbnails, since the whole
change is the colour values:

| Mode | Fill | Label | Rim |
|---|---|---|---|
| Light | `#FFBA27` | `#010101` | `#71510B`, 1.5 dp |
| Dark | `#F4B400` | `#010101` | none, fill clears the surface at 9.32 |


ksharma-xyz:
* **#2008** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/2008?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/2008?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #2011, opened 2026-09-20
docs: declare voice audio now that the mic flag is on

ksharma-xyz:
## What

`ai_search_input_enabled` is on for 1.27, so voice input reaches riders. That is the trigger
judgement call 2 in `docs/STORE_DATA_DECLARATIONS.md` was written to catch: the microphone
answer on both store forms was true only while the flag was off. Updates the declarations
doc to match, and adds voice search to the 1.27 store copy, which described the release as
if the feature were still switched off. Documentation only, no code.

## Changes

`docs/STORE_DATA_DECLARATIONS.md`:

- Microphone moves out of "read on the device and never transmitted" and into the table of
  what the app sends, naming the platform recogniser as the recipient.
- Play Data safety gains Audio / Voice or sound recordings: collected but **not** shared,
  processed ephemerally, App functionality, optional. Recorded from the submitted form. The
  row is removed from the not-declared table.
- Apple App Privacy gains User Content / Audio Data under Data Not Linked to You. The
  section header moves from eight types to nine.
- The "nothing is shared" statement stays as it was, and points at judgement call 2 for the
  one row where that distinction took an argument.
- Judgement call 2 is rewritten from "the answer expires when a flag flips" to the current
  state: KRAIL never receives the audio, both platforms prefer on-device transcription and
  fall back, and the not-shared answer rests on the transfer being rider-initiated and, on
  Android, going to a system service on the same device. The open product decision, requiring
  on-device recognition rather than preferring it, is recorded as open rather than decided.
- The Apple Device ID row is corrected to Analytics and App Functionality, read off the
  published form rather than an earlier state of it.
- The advertising ID entry records why the Play declaration answers Yes while the app
  removes the permission, and names the four GMS libraries that declare it.

`docs/release-notes/v1.27.0.md`:

- Voice search added as the lead item on both stores.
- Android copy re-cut to fit the 500 character limit with the new item in it, now 476.
- iOS list renumbered, now 900 characters.

## Testing

- No code changed, so no test task applies.
- Store copy character counts checked against both limits.
- `scripts/check_stale_assumptions.py` green.
- Checked the diff for arrows, middle dots and em dashes: none added.


ksharma-xyz:
* **#2011** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/2011?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/2011?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #2012, opened 2026-09-20
fix(ci): scope the GITHUB_TOKEN in the iOS test and nightly workflows

ksharma-xyz:
## What

Adds a workflow-level `permissions` block to `ios-unit-tests.yml` and `maestro-nightly.yml`.
Both ran with the repository default token scope, which CodeQL flagged as alerts 1, 3 and 4
under `actions/missing-workflow-permissions`. Those are the only three open code scanning
alerts, so this clears the list.

## Changes

- `.github/workflows/ios-unit-tests.yml` and `.github/workflows/maestro-nightly.yml` each
  declare `contents: read` and `packages: read` at workflow level, covering all three flagged
  jobs: `ios-unit-tests`, and `android` and `ios` in the nightly.
- `packages: read` is not optional here, and the block carries a comment saying why.
  `settings.gradle.kts` resolves the KRAIL-API-PROTO sources JAR from
  `maven.pkg.github.com` using `GITHUB_TOKEN`. Narrowing to `contents: read` alone would make
  Gradle fail to resolve it, which is why both workflows pass the token into Gradle in the
  first place.
- Declared at workflow level rather than per job, matching `build.yml` and
  `lint-workflows.yml`.

No job in either file writes anything back. They check out, build, run tests or flows, and
upload artifacts. `actions/upload-artifact` uses its own runtime token and does not need a
scope here.

## Testing

- `actionlint -shellcheck=` clean on both files. That is the exact invocation
  `lint-workflows.yml` uses, which disables the shellcheck integration while keeping the
  native checks.
- Both files parsed and the resulting `permissions` mapping confirmed on each.
- The real proof is the CI run on this PR: `ios-unit-tests` exercises the narrowed token
  against the GitHub Packages resolution described above.


ksharma-xyz:
* **#2012** <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/2012?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ksharma-xyz/KRAIL/2012?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

